PR #226, opened 2023-10-18
feat(market-params): use optional calldata

### LINKED ISSUE #211: [protocol review] Identify markets with `marketParams` instead of `Id`
Currently markets are identified by their `Id`. This is not optimal and requires to call `idToMarket` which was not designed for that purpose. See `supplyQueue` and `withdrawQueue` for example.
This potential optimization could save something like 10-15k per market: 5 * 2100 for cold SLOAD, plus (warm) call to Morpho.
Note that `reallocate` also works in this manner.

Because there will be a front end provided for vaults, this would make sense to do this optimization because it can be abstracted away for the users.

The problem (and I assume this is why it is done like that) is that to respect the ERC4626 interface, those entry points cannot take additional arguments. I see 3 ways around that issue:
1. memoize the arguments of the markets. This can be done by the curator when enabling a market, so it would not cost anything to the user. The expected gain is only the gas of a call to Morpho.
2. provide additional entrypoints that can be used by the front-end
3. Find a way to use additional calldata passed to the ERC4626 entry points as hints on what markets to use. Those hints would be verified to hash in the same way as the `supplyQueue` (resp. the `withdrawQueue`)

Rubilmax:
- Fixes #211

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -524,7 +524,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 nbMarkets = withdrawQueue.length;
 
         for (uint256 i; i < nbMarkets; ++i) {
-            assets += _supplyBalance(_marketParams(withdrawQueue[i]));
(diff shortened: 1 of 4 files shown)


MerlinEgalite:
Should we close?

PR #228, opened 2023-10-19
feat: add natspec about empty supply queue

### LINKED ISSUE #214: [protocol review] Allocators can effectively DOS the vault
Allocators have the power to set the supply Queue. In particular they are able to set an empty queue, or a queue with only markets that have reached the cap. In effect this means that all the supply goes to the idle liquidity.

I think this is an intended behavior, to have more granular control of the allocation. If this is the case, it should be documented above the `setSupplyQueue` function

Jean-Grimal:
Fixes #214 

Should I add a line explaining that in this case, the deposited assets stay idle until the next reallocation (done by allocators) ?


MerlinEgalite:
Closing this PR since we're going to remove the idle

MerlinEgalite:
Or maybe we should adapt the comment instead to make it more general.

MerlinEgalite:
Closing this PR since we're going to remove the idle

PR #259, opened 2023-10-23
Remove timelock expiration

### LINKED ISSUE #255: remove expiration mechanism
https://morpholabs.slack.com/archives/C02NE4SAE4B/p1698052613290089

MerlinEgalite:
Fixes #255


src/MetaMorpho.sol
```diff
@@ -153,12 +153,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// - there's no pending value;
     /// - the timelock has not elapsed since the pending value has been submitted;
     /// - the timelock has expired since the pending value has been submitted.
-    modifier withinTimelockWindow(uint256 submittedAt) {
+    modifier afterTimelock(uint256 submittedAt) {
```
QGarchery, line 156: We should also remove the comment just above this line about the timelock expiration

QGarchery (changes requested):

src/MetaMorpho.sol
MerlinEgalite: Good catch!

Rubilmax (approved):

MerlinEgalite (commented):
Update `README`

Rubilmax (approved):

Jean-Grimal (approved):

QGarchery (approved):
I still think that the behavior of changing ETA of pending values when changing the timelock is not expected (see #267). But let's consider that it is a separate issue.

LGTM otherwise

MerlinEgalite:
> I still think that the behavior of changing ETA of pending values when changing the timelock is unexpected (see #267). But let's consider that it is a separate issue.
> 
> LGTM otherwise

To me it should be in another PR on top

Rubilmax (approved):

Jean-Grimal (approved):

PR #281, opened 2023-10-31
Fix broken market

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/102

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -787,6 +787,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             Id id = withdrawQueue[i];
             MarketParams memory marketParams = _marketParams(id);
 
+            // Withdrawing from Morpho cannot fail in `_withdrawMorpho` because:
+            // 1. oracle.price() is never called (the vault doesn't borrow);
+            // 2. virtually accruing interest didn't fail in `_withdrawable`;
+            // 3. `_withdrawable` caps to the liquidity available on Morpho.
             remaining -= UtilsLib.min(_withdrawable(marketParams, id), remaining);
 
             if (remaining == 0) return 0;
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -485,11 +485,17 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's
+    /// IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts. The allocator
+    /// role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
     function maxWithdraw(address owner) public view override(IERC4626, ERC4626) returns (uint256 assets) {
         (assets,,) = _maxWithdraw(owner);
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's
+    /// IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts. The allocator
+    /// role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
     function maxRedeem(address owner) public view override(IERC4626, ERC4626) returns (uint256) {
         (uint256 assets, uint256 newTotalSupply, uint256 newTotalAssets) = _maxWithdraw(owner);
 
@@ -497,6 +503,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's
+    /// IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts. The allocator
+    /// role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
     function deposit(uint256 assets, address receiver) public override(IERC4626, ERC4626) returns (uint256 shares) {
         uint256 newTotalAssets = _accrueFee();
 
@@ -505,6 +514,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's
+    /// IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts. The allocator
+    /// role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
     function mint(uint256 shares, address receiver) public override(IERC4626, ERC4626) returns (uint256 assets) {
         uint256 newTotalAssets = _accrueFee();
 
@@ -541,11 +553,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
-    /// @dev Warning: Can revert. This goes against EIP4626. If a market is broken on Morpho `_supplyAssetsBalance` can
-    /// revert
-    /// as well as `totalAssets`. The allocator can always remove the broken market from `withdrawQueue` if the
-    /// situation
-    /// persists and unlock the vault.
+    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's
+    /// IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts. The allocator
+    /// role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
     function totalAssets() public view override(IERC4626, ERC4626) returns (uint256 assets) {
         for (uint256 i; i < withdrawQueue.length; ++i) {
             assets += _supplyAssetsBalance(_marketParams(withdrawQueue[i]));
@@ -563,7 +573,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @dev Returns the maximum amount of asset (`assets`) that the `owner` can withdraw from the vault, as well as the
     /// new vault's total supply (`newTotalSupply`) and total assets (`newTotalAssets`).
-    /// @dev Warning: Can revert.
     function _maxWithdraw(address owner)
         internal
         view
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -301,6 +301,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @notice Removing a market requires the vault to have 0 supply on it; but anyone can supply on behalf of the
     /// vault so the call to `sortWithdrawQueue` can be griefed by a frontrun. To circumvent this, the allocator can
     /// simply bundle a reallocation that withdraws max from this market with a call to `sortWithdrawQueue`.
+    /// @dev Should not revert.
     /// @param indexes The indexes of each market in the previous withdraw queue, in the new withdraw queue's order.
     function sortWithdrawQueue(uint256[] calldata indexes) external onlyAllocatorRole {
         uint256 newLength = indexes.length;
@@ -327,12 +328,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (!seen[i]) {
                 Id id = withdrawQueue[i];
 
-                // The following call cannot revert.
-                uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+                if (config[id].cap != 0) revert ErrorsLib.MissingMarket(id);
 
-                if (supplyShares != 0 || config[id].cap != 0) {
-                    revert ErrorsLib.MissingMarket(id);
-                }
+                bool success;
+                try this.expectedSupplyAssets(_marketParams(id)) {
+                    success = true;
+                } catch {}
+
+                if (success && MORPHO.supplyShares(id, address(this)) != 0) revert ErrorsLib.MissingMarket(id);
 
                 delete config[id].withdrawRank;
             }
@@ -470,22 +473,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /* MORPHO GETTERS */
 
     /// @notice Returns the expected supply asset balance of the vault on the market defined by `marketParams`.
-    function getExpectedSupplyBalance(MarketParams memory marketParams) external view returns (uint256) {
+    function expectedSupplyAssets(MarketParams memory marketParams) external view returns (uint256) {
         return MORPHO.expectedSupplyBalance(marketParams, address(this));
-    }
-
-    /// @notice Returns the expected balances of the market defined by `marketParams`.
-    function getExpectedMarketBalances(MarketParams memory marketParams)
-        external
-        view
-        returns (
-            uint256 totalSupplyAssets,
-            uint256 totalSupplyShares,
-            uint256 totalBorrowAssets,
-            uint256 totalBorrowShares
-        )
-    {
-        return MORPHO.expectedMarketBalances(marketParams);
     }
 
     /* ERC4626 (PUBLIC) */
@@ -552,10 +541,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: Can revert. This goes against EIP4626. If a market is broken on Morpho `_supplyBalance` can revert
+    /// as well as `totalAssets`. The allocator can always remove the broken market from `withdrawQueue` if the
+    /// situation
+    /// persists and unlock the vault.
     function totalAssets() public view override(IERC4626, ERC4626) returns (uint256 assets) {
         for (uint256 i; i < withdrawQueue.length; ++i) {
-            uint256 supplyAssets = _simulateSupplyBalance(_marketParams(withdrawQueue[i]));
-            assets += supplyAssets;
+            assets += _supplyBalance(_marketParams(withdrawQueue[i]));
         }
 
         assets += idle;
@@ -570,6 +562,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @dev Returns the maximum amount of asset (`assets`) that the `owner` can withdraw from the vault, as well as the
     /// new vault's total supply (`newTotalSupply`) and total assets (`newTotalAssets`).
+    /// @dev Warning: Can revert.
     function _maxWithdraw(address owner)
         internal
         view
@@ -663,16 +656,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         return MORPHO.expectedSupplyBalance(marketParams, address(this));
     }
 
-    /// @dev Returns the vault's assets balance the market defined by `marketParams`.
-    /// @dev To use when the function must not revert.
-    function _simulateSupplyBalance(MarketParams memory marketParams) internal view returns (uint256) {
-        try this.getExpectedSupplyBalance(marketParams) returns (uint256 balance) {
-            return balance;
-        } catch {
-            return 0;
-        }
-    }
-
     /// @dev Reverts if `newTimelock` is not within the bounds.
     function _checkTimelockBounds(uint256 newTimelock) internal pure {
         if (newTimelock > ConstantsLib.MAX_TIMELOCK) revert ErrorsLib.AboveMaxTimelock();
@@ -747,8 +730,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 toSupply = UtilsLib.min(_suppliable(marketParams, id), assets);
 
             if (toSupply > 0) {
-                MORPHO.supply(marketParams, toSupply, 0, address(this), hex"");
-                assets -= toSupply;
+                // Using try/catch to skip markets that revert.
+                try MORPHO.supply(marketParams, toSupply, 0, address(this), hex"") {
+                    assets -= toSupply;
+                } catch {}
             }
 
             if (assets == 0) return;
@@ -771,8 +756,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 toWithdraw = UtilsLib.min(_withdrawable(marketParams, id), remaining);
 
             if (toWithdraw > 0) {
-                MORPHO.withdraw(marketParams, toWithdraw, 0, address(this), address(this));
-                remaining -= toWithdraw;
+                // Using try/catch to skip markets that revert.
+                try MORPHO.withdraw(marketParams, toWithdraw, 0, address(this), address(this)) {
+                    remaining -= toWithdraw;
+                } catch {}
             }
 
             if (remaining == 0) return 0;
@@ -780,6 +767,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @dev Simulates a withdraw of `assets` from the idle liquidity and Morpho if necessary.
+    /// @dev Warning: Can revert.
     /// @return remaining The assets left to be withdrawn.
     function _simulateWithdrawMorpho(uint256 assets) internal view returns (uint256 remaining) {
         (remaining,) = _withdrawIdle(assets);
@@ -790,7 +778,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             Id id = withdrawQueue[i];
             MarketParams memory marketParams = _marketParams(id);
 
-            remaining -= UtilsLib.min(_simulateWithdrawable(marketParams, id), remaining);
+            remaining -= UtilsLib.min(_withdrawable(marketParams, id), remaining);
 
             if (remaining == 0) return 0;
         }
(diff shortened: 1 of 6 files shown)


src/MetaMorpho.sol
```diff
@@ -464,6 +466,32 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         emit EventsLib.TransferRewards(_msgSender(), rewardsRecipient, token, amount);
     }
 
+    /* MORPHO GETTERS */
+
+    /// @notice Returns the expected supply shares of the vault on the given market `id`.
+    function getSupplyShares(Id id) external view returns (uint256) {
```
MerlinEgalite, line 472: needed for the try catch

src/MetaMorpho.sol
```diff
@@ -638,6 +667,28 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         return MORPHO.expectedSupplyBalance(marketParams, address(this));
     }
 
+    /// @dev Returns the vault's assets balance the market defined by `marketParams`.
+    /// @dev To use when the function must not revert.
+    /// @return The vault's balance on the market defined by `marketParams`.
+    /// @return Whether the call succeeded or not. This information is needed when computing the suppliable amount.
+    function _simulateSupplyBalance(MarketParams memory marketParams) internal view returns (uint256, bool) {
```
MerlinEgalite, line 674: should this named differently?

src/MetaMorpho.sol
```diff
@@ -778,25 +830,39 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+        (uint256 supplyAssets, bool success) = _simulateSupplyBalance(marketParams);
+
+        if (success) return supplyCap.zeroFloorSub(supplyAssets);
+        else return 0;
     }
 
     /// @dev Returns the withdrawable amount of assets from the market defined by `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
+    /// @dev Must not revert.
     function _withdrawable(MarketParams memory marketParams, Id id) internal view returns (uint256) {
```
MerlinEgalite, line 847: ugly

src/MetaMorpho.sol
```diff
@@ -638,6 +667,29 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+        } catch {
+            return (0, false);
+        }
+    }
+
+    /// @dev Returns the vault's share balance the market defined by `id`.
+    /// @dev To use when the function must not revert.
+    /// @return Whether the call succeeded or not.
+    function _simulateSupplyShares(Id id) internal view returns (uint256, bool) {
+        try this.getSupplyShares(id) returns (uint256 shares) {
```
MerlinEgalite, line 686: perhaps we'd like to do a delegate call instead of a call to preserve the correct context

src/MetaMorpho.sol
```diff
@@ -712,10 +747,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 toSupply = UtilsLib.min(_suppliable(marketParams, id), assets);
 
             if (toSupply > 0) {
-                // Using try/catch to skip markets that revert.
-                try MORPHO.supply(marketParams, toSupply, 0, address(this), hex"") {
-                    assets -= toSupply;
-                } catch {}
+                MORPHO.supply(marketParams, toSupply, 0, address(this), hex"");
+                assets -= toSupply;
```
Jean-Grimal, line 751: So we assume that allocators allocators remove broken markets from the supply queue ? (I'm fine with it btw)

src/MetaMorpho.sol
```diff
@@ -738,19 +771,17 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 toWithdraw = UtilsLib.min(_withdrawable(marketParams, id), remaining);
 
             if (toWithdraw > 0) {
-                // Using try/catch to skip markets that revert.
-                try MORPHO.withdraw(marketParams, toWithdraw, 0, address(this), address(this)) {
-                    remaining -= toWithdraw;
-                } catch {}
+                MORPHO.withdraw(marketParams, toWithdraw, 0, address(this), address(this));
+                remaining -= toWithdraw;
```
Jean-Grimal, line 775: We should maybe keep the try/catch here as broken markets can't be removed from the withdraw queue, and broken markets might not always be at the end of the withdraw queue

src/MetaMorpho.sol
```diff
@@ -799,6 +827,26 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         return UtilsLib.min(supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares), availableLiquidity);
     }
 
+    /// @dev Returns the withdrawable amount of assets from the market defined by `marketParams`.
+    /// @dev Assumes that the inputs `marketParams` and `id` match.
+    /// @dev Must not revert.
+    function _simulateWithdrawable(MarketParams memory marketParams, Id id) internal view returns (uint256) {
```
Jean-Grimal, line 833: Shouldn't we just replace `_withdrawable` with this ?

src/MetaMorpho.sol
```diff
@@ -327,7 +327,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (!seen[i]) {
                 Id id = withdrawQueue[i];
 
-                if (MORPHO.supplyShares(id, address(this)) != 0 || config[id].cap != 0) {
+                // The following call cannot revert.
```
Rubilmax, line 330: Wdyt about specifying in the natspecs instead:
`/// @dev Should not revert.`

src/MetaMorpho.sol
```diff
@@ -464,6 +467,27 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         emit EventsLib.TransferRewards(_msgSender(), rewardsRecipient, token, amount);
     }
 
+    /* MORPHO GETTERS */
+
+    /// @notice Returns the expected supply asset balance of the vault on the market defined by `marketParams`.
+    function getExpectedSupplyBalance(MarketParams memory marketParams) external view returns (uint256) {
```
Rubilmax, line 473:
```suggestion
    function expectedSupplyAssets(MarketParams memory marketParams) external view returns (uint256) {
```

Rubilmax (commented):
It doesn't correspond to the minimal changes I suggest right? Is it expected or is it simply because you made this PR before my comment?.

MerlinEgalite:
> It doesn't correspond to the minimal changes I suggest right? Is it expected or is it simply because you made this PR before my comment?

I did not have the time to apply any more changes. 

I'm not sure about what is part or not of your proposition since you wrote several messages.

src/MetaMorpho.sol
MerlinEgalite: we can do both I think

src/MetaMorpho.sol
```diff
@@ -530,7 +554,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IERC4626
     function totalAssets() public view override(IERC4626, ERC4626) returns (uint256 assets) {
         for (uint256 i; i < withdrawQueue.length; ++i) {
-            assets += _supplyBalance(_marketParams(withdrawQueue[i]));
+            uint256 supplyAssets = _simulateSupplyBalance(_marketParams(withdrawQueue[i]));
```
MerlinEgalite, line 557: If I understand your point @Rubilmax you suggest to revert on `totalAssets` and too bad for the ERC4626, right?

src/MetaMorpho.sol
MerlinEgalite: Yes I think this is a good assumption

src/MetaMorpho.sol
MerlinEgalite: Yes they can since we do a try/catch now!

src/MetaMorpho.sol
MerlinEgalite: It costs more gas though so I'm not sure.

But we could argue that it removes a function

src/MetaMorpho.sol
Rubilmax: Correct! I only suggest to change `sortWithdrawQueue`

src/MetaMorpho.sol
Jean-Grimal: Apparently the try/catch is removed here

src/MetaMorpho.sol
```diff
@@ -464,6 +470,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         emit EventsLib.TransferRewards(_msgSender(), rewardsRecipient, token, amount);
     }
 
+    /* MORPHO GETTERS */
+
+    /// @notice Returns the expected supply asset balance of the vault on the market defined by `marketParams`.
+    function expectedSupplyAssets(MarketParams memory marketParams) external view returns (uint256) {
+        return MORPHO.expectedSupplyBalance(marketParams, address(this));
```
MerlinEgalite, line 477: Maybe we should point to Morpho Blue branch with the renaming.

In this case we should rename `_supplyBalance` as well

src/MetaMorpho.sol
```diff
@@ -327,9 +328,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-                    revert ErrorsLib.MissingMarket(id);
-                }
+                if (config[id].cap != 0) revert ErrorsLib.MissingMarket(id);
+
+                bool success;
+                try this.expectedSupplyAssets(_marketParams(id)) {
+                    success = true;
+                } catch {}
+
+                if (success && MORPHO.supplyShares(id, address(this)) != 0) revert ErrorsLib.MissingMarket(id);
```
Rubilmax, line 338:
```suggestion
                try this.expectedSupplyAssets(_marketParams(id)) returns (uint256 supplyAssets) {
                  if (supplyAssets != 0) revert ErrorsLib.MissingMarket(id);
                } catch {}
```

src/MetaMorpho.sol
```diff
@@ -787,6 +803,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @dev Returns the withdrawable amount of assets from the market defined by `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
+    /// @dev Warning: Can revert.
```
Rubilmax, line 806: If we write this in internal functions, shouldn't we also write it for all functions that call them?

src/MetaMorpho.sol
```diff
@@ -748,9 +766,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
-    /// @dev Fakes a withdraw of `assets` from the idle liquidity and Morpho if necessary.
+    /// @dev Simulates a withdraw of `assets` from the idle liquidity and Morpho if necessary.
+    /// @dev Warning: Can revert.
     /// @return remaining The assets left to be withdrawn.
-    function _staticWithdrawMorpho(uint256 assets) internal view returns (uint256 remaining) {
+    function _simulateWithdrawMorpho(uint256 assets) internal view returns (uint256 remaining) {
```
Rubilmax, line 772: Why is this change in this PR? Shouldn't it be based on #280 ?

src/MetaMorpho.sol
```diff
@@ -759,10 +778,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             Id id = withdrawQueue[i];
             MarketParams memory marketParams = _marketParams(id);
 
-            // The vault withdrawing from Morpho cannot fail because:
-            // 1. oracle.price() is never called (the vault doesn't borrow)
-            // 2. `_withdrawable` caps to the liquidity available on Morpho
-            // 3. virtually accruing interest didn't fail in `_withdrawable`
```
Rubilmax, line 765: Why is this removed? This comment is still correct

src/MetaMorpho.sol
```diff
@@ -528,6 +541,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev Warning: Can revert. This goes against EIP4626. If a market is broken on Morpho `_supplyBalance` can revert
+    /// as well as `totalAssets`. The allocator can always remove the broken market from `withdrawQueue` if the
+    /// situation
+    /// persists and unlock the vault.
```
Rubilmax, line 547:
```suggestion
    /// @dev Warning: May revert when virtually accruing interest on a market of the withdraw queue, due to the market's IRM. In such case, funds supplied to the market are lost as long as accruing interest reverts and the allocator role can remove the market from `withdrawQueue` to realize the loss and unlock the vault.
```

1. I don't like the naming "broken" markets because it is too vague
2. If we're adding this natspec to this function, we should add it to all functions that depend on it?

Rubilmax (changes requested):

src/MetaMorpho.sol
MerlinEgalite: I'll pull the `cantina-review` if you want

src/MetaMorpho.sol
MerlinEgalite: why? `supplyShares` is more accurate in this case

src/MetaMorpho.sol
MerlinEgalite: 1. agree
2. Not possible because some are in the ERC4626 implementation

src/MetaMorpho.sol
MerlinEgalite: We can remove it

src/MetaMorpho.sol
MerlinEgalite: Actually there are some conflicts so it's mandatory anyway

src/MetaMorpho.sol
MerlinEgalite: [Related discussion](https://morpholabs.slack.com/archives/C02N7CSP91C/p1699019069613909) because there can have some implications if we to add `get` before getters

src/MetaMorpho.sol
Rubilmax: Should we still duplicate it for all functions that we override and that depend on `totalAssets`? I fear this comment gets lost because the integrator doesn't connect the dots

MerlinEgalite: Better to replicate, than nothing I'd say so yes let's do that.

src/MetaMorpho.sol
MerlinEgalite: I think it's not correct.

> // 3. virtually accruing interest didn't fail in `_withdrawable`

`_withdrawable` can fail because of the IRM.

src/MetaMorpho.sol
```diff
@@ -327,9 +328,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (!seen[i]) {
                 Id id = withdrawQueue[i];
 
-                if (MORPHO.supplyShares(id, address(this)) != 0 || config[id].cap != 0) {
-                    revert ErrorsLib.MissingMarket(id);
-                }
+                if (config[id].cap != 0) revert ErrorsLib.MissingMarket(id);
```
MerlinEgalite, line 331: we may want to log different error messages in fact

src/MetaMorpho.sol
Rubilmax: Yes, but that's not what is stated in this comment
The comment states that withdrawing from Morpho cannot fail, given the 3 assumptions
Perhaps you want to rephrase it

src/MetaMorpho.sol
MerlinEgalite: I added the comment on entry points and getters but not internal functions. I think this is the most important places

src/MetaMorpho.sol
MerlinEgalite: I don't get the point of this comment in fact. `_simulateWithdrawMorpho` is used in `_maxWithdraw` in turn used in `maxWithdraw` and `maxRedeem`.

So there's no withdraw involved.

Maybe we want to say the following:

> If `_simulateWithdrawMorpho` did not fail then a withdraw on Morpho won't fail.

It's true only atomically though.

Or am I missing something?

Rubilmax: The point of the comment is to justify that `_simulateWithdrawMorpho` accounts for what's expected, ie what will actually be accounted for by `_withdrawMorpho`

In `_withdrawMorpho` we try/catch so there may be a hidden revert in `_withdrawMorpho` that may not be accounted for in `_simulateWithdrawMorpho`
So the comment actually justifies that this case doesn't exist, given the 3 assumptions listed

src/MetaMorpho.sol
Rubilmax: But the following property holds: `supplyShares == 0 => supplyAssets == 0`
Reciprocally, `supplyAssets != 0 => supplyShares != 0`

So it's more optimized

src/MetaMorpho.sol
MerlinEgalite: To me the comment as it is now is unclear. And maybe we should move the comment in the Natspecs instead

src/MetaMorpho.sol
Rubilmax: Right, my main concern is that we're doing twice the CALL to `MORPHO.supplyShares`

The minimum refactoring is:
```solidity
try this.expectedSupplyAssets(_marketParams(id)) {
  if (MORPHO.supplyShares(id, address(this)) != 0) revert ErrorsLib.MissingMarket(id);
} catch {}
```

Then from this point, we have the choice to:
1. do nothing
2. change `expectedSupplyAssets` to `expectedSupplyBalance() returns (uint256 shares, uint256 assets)` and use the return value `shares`

src/MetaMorpho.sol
Rubilmax:
```solidity
// Withdrawing from Morpho cannot fail in `_withdrawMorpho` because:
// 1. oracle.price() is never called (the vault doesn't borrow)
// 2. virtually accruing interest didn't fail in `_withdrawable`
// 3. `_withdrawable` caps to the liquidity available on Morpho
```

Rubilmax: And if you want update the natspec of `_simulateWithdrawMorpho` to specify that it replicates the behavior of `_withdrawMorpho` and should account for the same things

Rubilmax (changes requested):

src/MetaMorpho.sol
MerlinEgalite: The function `expectedSupplyBalance` that you mentioned does not exist though...

Rubilmax: We can create it, without using `MorphoBalancesLib.expectedSupplyAssets` directly!

MerlinEgalite: > We can create it, without using MorphoBalancesLib.expectedSupplyAssets directly!

I'll do this option on top

Rubilmax: We could also simply change the spec and allow removing a market from the withdraw queue as soon as the nb of shares on this market doesn't correspond to a single atom of the underlying asset

The reason we're limiting this from the start was to guarantee lenders the possibility to exit the vault at any time, given the liquidity on underlying markets. If the vault doesn't have a single atom of underlying asset supplied on a market, who cares about being able to withdraw from it?

With accrued interest, one may argue that this possibly non-zero nb of shares may grow to a non-zero supply of assets. To what extent? Let's imagine that interest grows 10 times the volume of the underlying market: the non-zero supply of shares would correspond to a maximum of 10 atoms of assets. It's just dust, so who cares about being able to withdraw this liquidity? The allocator can always reallocate it btw.

src/libraries/ErrorsLib.sol
```diff
@@ -39,8 +39,11 @@ library ErrorsLib {
     /// @notice Thrown when market `id` is a duplicate in the new withdraw queue to set.
     error DuplicateMarket(Id id);
 
-    /// @notice Thrown when market `id` is missing in the new withdraw queue to set.
-    error InvalidMarketRemoval(Id id);
+    /// @notice Thrown when market `id` has a non-zero cap and is missing in the new withdraw queue to set.
+    error InvalidMarketRemovalNonZeroCap(Id id);
```
Rubilmax, line 43:
```suggestion
    error InvalidMarketRemovalNonZeroCap(Id indexed id);
```

src/libraries/ErrorsLib.sol
```diff
@@ -39,8 +39,11 @@ library ErrorsLib {
...
+
+    /// @notice Thrown when market `id` has a non-zero vault's supply and is missing in the new withdraw queue to set.
+    error InvalidMarketRemovalNonZeroSupply(Id id);
```
Rubilmax, line 46:
```suggestion
    error InvalidMarketRemovalNonZeroSupply(Id indexed id);
```

Rubilmax: Or this should be done in a dedicated PR, but we should index these in all events

Rubilmax (changes requested):

src/libraries/ErrorsLib.sol
MerlinEgalite: It's an error not an event

Rubilmax: lmao

Jean-Grimal (approved):

src/MetaMorpho.sol
```diff
@@ -302,6 +302,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @notice Removing a market requires the vault to have 0 supply on it; but anyone can supply on behalf of the
     /// vault so the call to `sortWithdrawQueue` can be griefed by a frontrun. To circumvent this, the allocator can
     /// simply bundle a reallocation that withdraws max from this market with a call to `sortWithdrawQueue`.
+    /// @dev Should not revert.
```
Rubilmax, line 305: After thinking about it, this comment alone does not make sense because we do want this function to revert for specific reasons

src/MetaMorpho.sol
MerlinEgalite: Good point indeed, I think this is more that you still get expose to some extent to a market. 

But I agree that this is a ridiculously small amount and would simplify the code...

@Jean-Grimal do you have a though?

src/MetaMorpho.sol
MerlinEgalite: Right, I'll remove it

src/MetaMorpho.sol
Jean-Grimal: What is a single atom ?

MerlinEgalite: This Screenshot 2023-11-06 at 18 44 45

Joke aside, I think what @Rubilmax means is that if the shares are worth 0 tokens, then we don't care about those remaining shares and we can skip the market

Rubilmax: The indivisible unit of an asset (e.g. 1 wei for ETH)

Jean-Grimal: I agree with @Rubilmax, I don't think removing markets with less than 1 atom worth of supply is a problem.
And actually I think in over 99% of cases, markets with less than 1 atom worth of supply will be markets with 0 shares.

Rubilmax: I agree iff #172 is merged

Rubilmax (approved):

Jean-Grimal (approved):

QGarchery (changes requested):
I'm strongly against the changes of this PR, because it gives some significant power to allocators to accomodate for a worst case scenario. Notably, the allocators are able to manipulate the value of the shares with the following attack:
1. put the biggest market with cap 0 at the end of the withdraw queue
2. call `sortWithdrawQueue` with a special amount of gas so that the last iteration reverts and removes the market
3. supply a lot

After these steps, the supply will eventually gain in value if the admins want to recover the funds of the  removed market. 

I think that if we want to remove a market we have to make it more explicit and give this power only to the owner or maybe the curator. Note that there are [other solutions](https://morpholabs.slack.com/archives/C02N7CSP91C/p1698914655382309?thread_ts=1698910412.624549&cid=C02N7CSP91C) that do not break our invariants (notably about the total amount of funds being in the withdraw queue)

src/MetaMorpho.sol
```diff
@@ -328,9 +328,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
 
-                if (MORPHO.supplyShares(id, address(this)) != 0 || config[id].cap != 0) {
-                    revert ErrorsLib.InvalidMarketRemoval(id);
-                }
+                if (config[id].cap != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroCap(id);
+
+                try this.expectedSupplyAssets(_marketParams(id)) returns (uint256 supplyAssets) {
+                    // Assumes that a market with zero supply assets (but possibly non-zero supply shares) is dust.
+                    if (supplyAssets != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroSupply(id);
+                } catch {}
```
Rubilmax, line 336:
```suggestion
                if (isAllocator[_msgSender()] && MORPHO.supplyShares(id) != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroSupply(id);
```

Rubilmax (changes requested):
Following latest discussion

src/MetaMorpho.sol
MerlinEgalite: isn't curator that should do it instead?

MerlinEgalite: Oh right, I had the opposite in my head. So indeed if it's the allocator that triggers the function we perform the check.

Don't we want to do the try/catch for curator/owner though? Else it's very easy to remove a market

Rubilmax: As stated by @QGarchery, the try/catch is vulnerable to OOG, so it wouldn't be a valid additional safety measure

Rubilmax:
> 2. call `sortWithdrawQueue` with a special amount of gas so that the last iteration reverts and removes the market

How can a specific amount of gas exist such that the EVM forwards 63/64 gas to the Morpho CALL, that it reverts but that 1/64 remaining gas is enough to remove the market from the new withdraw queue?
There may be proof that this amount exists / that it never does

src/MetaMorpho.sol
QGarchery: > How would an amount of total gas exist such that the call to Morpho reverts, considering the EVM forwards 63/64 of remaining gas to Morpho? @QGarchery

I'm not sure that it is feasible, but I think that the idea is to not have enough gas with 63/64 of the remaining gas to complete the morpho call, but have enough gas with 1/64 to complete the loop (which could potentially happen if you put the faulty market at the end)

Rubilmax:
Replaced by #312

PR #282, opened 2023-10-31
Replace the withdraw rank with a boolean

QGarchery:
Credits to Maxim from OZ for this finding.

The purpose is performance, wich should come mainly from the removal of the SSTORE in the withraw loop inside `reallocate`

### DIFF
--- a/src/interfaces/IMetaMorpho.sol
+++ b/src/interfaces/IMetaMorpho.sol
@@ -7,8 +7,8 @@ import {IERC4626} from "@openzeppelin/interfaces/IERC4626.sol";
 struct MarketConfig {
     /// @notice The maximum amount of assets that can be allocated to the market.
     uint192 cap;
-    /// @notice The rank of the market in the withdraw queue.
-    uint64 withdrawRank;
+    /// @notice Whether the market is in the withdraw queue.
+    bool inWithdrawQueue;
 }
 
 struct PendingUint192 {
@@ -50,7 +50,7 @@ interface IMetaMorpho is IERC4626 {
     function supplyQueueSize() external view returns (uint256);
     function withdrawQueue(uint256) external view returns (Id);
     function withdrawQueueSize() external view returns (uint256);
-    function config(Id) external view returns (uint192 cap, uint64 withdrawRank);
+    function config(Id) external view returns (uint192 cap, bool inWithdrawQueue);
 
     function idle() external view returns (uint256);
     function lastTotalAssets() external view returns (uint256);
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -318,9 +318,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             seen[prevIndex] = true;
 
             newWithdrawQueue[i] = id;
-
-            // Safe "unchecked" cast because i < currLength.
-            config[id].withdrawRank = uint64(i + 1);
         }
 
         for (uint256 i; i < currLength; ++i) {
@@ -331,7 +328,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                     revert ErrorsLib.MissingMarket(id);
                 }
 
-                delete config[id].withdrawRank;
+                config[id].inWithdrawQueue = false;
             }
         }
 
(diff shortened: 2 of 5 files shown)


MerlinEgalite (approved):
Seems to work indeed

Rubilmax (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -7,8 +7,8 @@ import {IERC4626} from "@openzeppelin/interfaces/IERC4626.sol";
 struct MarketConfig {
     /// @notice The maximum amount of assets that can be allocated to the market.
     uint192 cap;
-    /// @notice The rank of the market in the withdraw queue.
-    uint64 withdrawRank;
+    /// @notice Whether the market is in the withdraw queue.
+    bool inWithdrawQueue;
```
Jean-Grimal, line 11:
```suggestion
    bool enabled;
```
Actually enabled markets are markets in the withdraw queue. Could make sense to make it explicit with this naming

Jean-Grimal (changes requested):

src/interfaces/IMetaMorpho.sol
Rubilmax: Smart! I like it

MerlinEgalite (approved):

Jean-Grimal (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -50,7 +50,7 @@ interface IMetaMorpho is IERC4626 {
     function supplyQueueLength() external view returns (uint256);
     function withdrawQueue(uint256) external view returns (Id);
     function withdrawQueueLength() external view returns (uint256);
-    function config(Id) external view returns (uint192 cap, uint64 withdrawRank);
+    function config(Id) external view returns (uint192 cap, uint64 enabled);
```
Rubilmax, line 53:
```suggestion
    function config(Id) external view returns (uint192 cap, bool enabled);
```

src/interfaces/IMetaMorpho.sol
```diff
@@ -7,8 +7,8 @@ import {IERC4626} from "@openzeppelin/interfaces/IERC4626.sol";
 struct MarketConfig {
     /// @notice The maximum amount of assets that can be allocated to the market.
     uint192 cap;
```
Rubilmax, line 9: Then we could extend this type up to `uint248`

src/interfaces/IMetaMorpho.sol
```diff
@@ -7,8 +7,8 @@ import {IERC4626} from "@openzeppelin/interfaces/IERC4626.sol";
...
-    /// @notice The rank of the market in the withdraw queue.
-    uint64 withdrawRank;
+    /// @notice Whether the market is in the withdraw queue.
+    bool enabled;
```
Rubilmax, line 11: We may also want to switch the order to have the masked data part be the less used
1. the less used in the code is clearly `enabled`
2. but currently, is it `enabled` that needs to be masked to be read or is it `cap`?

src/interfaces/IMetaMorpho.sol
QGarchery: :sweat_smile:

src/interfaces/IMetaMorpho.sol
QGarchery: To get any value in a slot you shift it at the beginning of the word, and mask it afterwards. You can optimize this by:
1. not shifting when the value is already the beginning of the slot
2. not mask it when the value is already at the end of the slot

It seems like solidity does not do optimization 2, see [this PR](https://github.com/morpho-labs/morpho-sandbox/pull/16). So in theory it would cost less to put the most used at the beginning of the struct/slot.

But I think that this is such a small optimization that we should not bother with it at all

src/interfaces/IMetaMorpho.sol
QGarchery: I don't think we should do it because pending cap is used as an `uint192` everywhere. And we can't change it easily because it is also used in `PendingUint192`

src/interfaces/IMetaMorpho.sol
Rubilmax: Thanks for investigating!

Rubilmax (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):

PR #256, opened 2023-10-23
Remove timelock for fee

MerlinEgalite:
Fixes #249

PR #292, opened 2023-11-02
refactor(timelock): change submittedAt to validAt

### LINKED ISSUE #267: Unpredictable timelock mechanism
The issue highlighted in https://github.com/cantinasec/review-morpho-blue-1/pull/68#discussion_r1367673403 seems to be fixed with #259. But actually this is a deeper issue where timelock of pending values are variable which can be unexpected. Consider the following scenario:
1. Let's say the timelock is initially of 1 week, mark the current time as `t`
2. People agree to change the timelock to 1 day, a pending value is submitted at `t`, with the intent that it is effective at `t + 7`
3. 3 days later, at `t + 3`, people agree to increase the fee, with the intent that it is effective at `t + 10`
4. At `t + 7` the timelock is accepted
5. After that (starting at `t + 7` and beyond) the fee can be increased

I think that we should change the `submittedAt` field of pending values to a `validFrom` field that would store `block.timestamp + timelock` and the moment where the pending value is set

Rubilmax:
- Fixes #267

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -208,11 +208,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newTimelock > timelock) {
             _setTimelock(newTimelock);
         } else {
-            pendingTimelock.update(
-                // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
-                uint192(newTimelock),
-                timelock
-            );
+            // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
+            pendingTimelock.update(uint192(newTimelock), timelock);
 
             emit EventsLib.SubmitTimelock(newTimelock);
         }
@@ -228,11 +225,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newFee < fee) {
             _setFee(newFee);
         } else {
-            pendingFee.update(
-                // Safe "unchecked" cast because newFee <= MAX_FEE.
-                uint192(newFee),
-                timelock
-            );
+            // Safe "unchecked" cast because newFee <= MAX_FEE.
+            pendingFee.update(uint192(newFee), timelock);
 
             emit EventsLib.SubmitFee(newFee);
         }
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -2,11 +2,10 @@
 pragma solidity 0.8.21;
 
 import {IMorphoMarketParams} from "./interfaces/IMorphoMarketParams.sol";
-import {
-    IMetaMorpho, MarketConfig, PendingUint192, PendingAddress, MarketAllocation
-} from "./interfaces/IMetaMorpho.sol";
+import {IMetaMorpho, MarketConfig, MarketAllocation} from "./interfaces/IMetaMorpho.sol";
 import {Id, MarketParams, Market, IMorpho} from "@morpho-blue/interfaces/IMorpho.sol";
 
+import {PendingUint192, PendingAddress, PendingLib} from "./libraries/PendingLib.sol";
 import {ConstantsLib} from "./libraries/ConstantsLib.sol";
 import {ErrorsLib} from "./libraries/ErrorsLib.sol";
 import {EventsLib} from "./libraries/EventsLib.sol";
@@ -36,6 +35,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     using SharesMathLib for uint256;
     using MorphoBalancesLib for IMorpho;
     using MarketParamsLib for MarketParams;
+    using PendingLib for PendingUint192;
+    using PendingLib for PendingAddress;
 
     /* IMMUTABLES */
 
@@ -152,9 +153,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @dev Reverts if:
     /// - there's no pending value;
     /// - the timelock has not elapsed since the pending value has been submitted.
-    modifier afterTimelock(uint256 submittedAt) {
-        if (submittedAt == 0) revert ErrorsLib.NoPendingValue();
-        if (block.timestamp < submittedAt + timelock) revert ErrorsLib.TimelockNotElapsed();
+    modifier afterTimelock(uint256 validAt) {
+        if (validAt == 0) revert ErrorsLib.NoPendingValue();
+        if (block.timestamp < validAt) revert ErrorsLib.TimelockNotElapsed();
 
         _;
     }
@@ -198,8 +199,11 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newTimelock > timelock) {
             _setTimelock(newTimelock);
         } else {
-            // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
-            pendingTimelock = PendingUint192(uint192(newTimelock), uint64(block.timestamp));
+            pendingTimelock.update(
+                // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
+                uint192(newTimelock),
+                timelock
+            );
 
             emit EventsLib.SubmitTimelock(newTimelock);
         }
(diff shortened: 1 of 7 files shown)


Jean-Grimal (approved):
I'd rather stay with the current version though

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
...
+}
+
+struct PendingAddress {
+    /// @notice The pending value to set.
+    address value;
+    /// @notice The timestamp at which the value was submitted.
+    uint96 validAt;
+}
+
+/// @title ConstantsLib
```
MerlinEgalite, line 18:
```suggestion
/// @title PendingLib
```

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+struct PendingUint192 {
+    /// @notice The pending value to set.
+    uint192 value;
+    /// @notice The timestamp at which the value was submitted.
```
MerlinEgalite, line 7:
```suggestion
    /// @notice The timestamp at which the pending value becomes valid.
```

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
...
+    uint64 validAt;
+}
+
+struct PendingAddress {
+    /// @notice The pending value to set.
+    address value;
+    /// @notice The timestamp at which the value was submitted.
```
MerlinEgalite, line 14:
```suggestion
    /// @notice The timestamp at which the pending value becomes valid.
```

MerlinEgalite (changes requested):

Rubilmax:
> I'd rather stay with the current version though

But there's a bug with it, see the associated issue
What specifically don't you like about the new version? We can modify things to find a compromise

Jean-Grimal:
> > I'd rather stay with the current version though
> 
> But there's a bug with it, see the associated issue What specifically don't you like about the new version? We can modify things to find a compromise

Oh yes sorry I was mistaken. I understand the issue now

MerlinEgalite (approved):

Jean-Grimal (approved):

MerlinEgalite (approved):

MerlinEgalite:
@Rubilmax can you update this PR please?

MerlinEgalite (approved):

MerlinEgalite (changes requested):
I think the events haven't been updated though

src/MetaMorpho.sol
```diff
@@ -201,8 +202,11 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newTimelock > timelock) {
             _setTimelock(newTimelock);
         } else {
-            // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
-            pendingTimelock = PendingUint192(uint192(newTimelock), uint64(block.timestamp));
+            pendingTimelock.update(
+                // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
+                uint192(newTimelock),
+                timelock
+            );
```
QGarchery, line 209:
```suggestion
            // Safe "unchecked" cast because newTimelock <= MAX_TIMELOCK.
            pendingTimelock.update(uint192(newTimelock), timelock);
```
just to make it more readable (personal opinion)

MerlinEgalite (approved):

src/interfaces/IMetaMorpho.sol
QGarchery, line 1: There will be many conflicts with #272, maybe base this PR on top ?

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
...
+        // Safe "unchecked" cast because timelock <= MAX_TIMELOCK.
+        pending.validAt = uint64(block.timestamp) + uint64(timelock);
+    }
+
+    /// @dev Updates `pending`'s value to `newValue` and its corresponding `validAt` timestamp.
+    /// @dev Assumes `timelock` <= `MAX_TIMELOCK`.
+    function update(PendingAddress storage pending, address newValue, uint256 timelock) internal {
+        pending.value = newValue;
+        // Safe "unchecked" cast because timelock <= MAX_TIMELOCK.
+        pending.validAt = uint64(block.timestamp) + uint64(timelock);
```
QGarchery, line 36:
```suggestion
        pending.validAt = uint96(block.timestamp) + uint96(timelock);
```
because otherwise there will be another implicit cast to `uint96` anyway

QGarchery (commented):
Adding 6 tests :muscle:

A few minor things to change but I like this implementation !

Rubilmax:
> Adding 6 tests 💪
> 
> A few minor suggestions but I like this implementation !

4 of them will actually be removed in #256

src/MetaMorpho.sol
QGarchery, line 1: The target branch of this PR should be changed to `refactor/interfaces-3`

src/libraries/PendingLib.sol
QGarchery: This is not resolved right ?

QGarchery (changes requested):

src/libraries/PendingLib.sol
Rubilmax: Yes it is, by simply storing `validAt` as `uint64` ^^

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
...
```
MerlinEgalite, line 36:
```suggestion
        pending.validAt = uint64(block.timestamp + timelock);
```

Isn't possible?

MerlinEgalite (approved):

src/libraries/PendingLib.sol
```diff
@@ -0,0 +1,38 @@
...
+        pending.value = newValue;
+        // Safe "unchecked" cast because timelock <= MAX_TIMELOCK.
+        pending.validAt = uint64(block.timestamp + timelock);
+    }
+
+    /// @dev Updates `pending`'s value to `newValue` and its corresponding `validAt` timestamp.
+    /// @dev Assumes `timelock` <= `MAX_TIMELOCK`.
+    function update(PendingAddress storage pending, address newValue, uint256 timelock) internal {
+        pending.value = newValue;
+        // Safe "unchecked" cast because timelock <= MAX_TIMELOCK.
```
StErMi, line 35: You are referencing something that is not defined in this library.

Rubilmax: I think there are other places like this. Should I:
```suggestion
        // Safe "unchecked" cast because timelock <= ConstantsLib.MAX_TIMELOCK.
```

@morpho-org/onchain-team ?

MerlinEgalite: I don't think it's worth it tbh

StErMi:
@MerlinEgalite just to recap the changes

- the owner now cannot directly influence the timelock of existing pending proposals (already scheduled) by calling `submitTimelock(newTimelock)` where `newTimelock > timelock`
- the owner can still influence every pending proposal by executing `submit*`. He can also pass the very same already proposed value. 
- the owner can anyway revoke the pending proposal and propose a new one (not in this PR, but there's another PR that makes the `owner` part of the guardian role) 

PS: is there a PR that does not allow setting the same values for the current active pending proposals?
TLDR the main difference is that the duration of the timelock is stored in the pending proposal when it's created and `submitTimelock(newTimelock)` cannot directly influence an already scheduled pending proposal even if the `timelock` is instantly changed or `pendingTimelock` is approved.

MerlinEgalite:
> * the owner now cannot directly influence the timelock of existing pending proposals (already scheduled) by calling `submitTimelock(newTimelock)` where `newTimelock > timelock`
> * the owner can still influence every pending proposal by executing `submit*`. He can also pass the very same already proposed value.
> * the owner can anyway revoke the pending proposal and propose a new one (not in this PR, but there's another PR that makes the `owner` part of the guardian role)

This is exact in the current state of this PR.

> PS: is there a PR that does not allow setting the same values for the current active pending proposals? TLDR the main difference is that the duration of the timelock is stored in the pending proposal when it's created and `submitTimelock(newTimelock)` cannot directly influence an already scheduled pending proposal even if the `timelock` is instantly changed or `pendingTimelock` is approved.

yes there's such PR here: https://github.com/morpho-org/metamorpho/pull/295

PR #293, opened 2023-11-02
Rename `MissingMarket`

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/100


Rubilmax (approved):
I still don't think it's necessary but perhaps having a fresh eye on the codebase would help decide: ask @PaulFrambot ??

Jean-Grimal (approved):

MerlinEgalite:
> I still don't think it's necessary but perhaps having a fresh eye on the codebase would help decide: ask @PaulFrambot ??

I think the implication is very low so we don't need to trigger him for that

PR #298, opened 2023-11-02
fix(reallocate): prevent slippage

### LINKED ISSUE #296: Add slippage protection to `reallocate`

Rubilmax:
- Fixes #296 

This PR protects the allocator from bearing slippage by simply not allowing to withdraw/supply shares. The allocator can still withdraw all from a market by passing `type(uint256).max`

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -351,19 +351,22 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 totalWithdrawn;
         for (uint256 i; i < withdrawn.length; ++i) {
             MarketAllocation memory allocation = withdrawn[i];
+            Id id = allocation.marketParams.id();
 
             if (allocation.marketParams.loanToken != asset()) {
-                revert ErrorsLib.InconsistentAsset(allocation.marketParams.id());
+                revert ErrorsLib.InconsistentAsset(id);
             }
 
             // Guarantees that unknown frontrunning donations can be withdrawn, in order to disable a market.
-            if (allocation.shares == type(uint256).max) {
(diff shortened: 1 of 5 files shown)


MerlinEgalite (approved):
Back to the good old days?

src/MetaMorpho.sol
```diff
@@ -351,19 +351,22 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 totalWithdrawn;
         for (uint256 i; i < withdrawn.length; ++i) {
             MarketAllocation memory allocation = withdrawn[i];
+            Id id = allocation.marketParams.id();
```
Rubilmax, line 354: I'm not sure we should do this, because this is only used once in the code (once in each branch of the code)

MerlinEgalite: I think the code is cleaner like that, I would keep it

Jean-Grimal (approved):

MerlinEgalite:
replaced by #307

Rubilmax:
> replaced by #307

It is based on it! The 2 PRs don't address the same issue

MerlinEgalite:
I you're right, can you update it then?

src/interfaces/IMetaMorpho.sol
```diff
@@ -31,8 +31,6 @@ struct MarketAllocation {
     MarketParams marketParams;
     /// @notice The amount of assets to allocate.
     uint256 assets;
-    /// @notice The amount of shares to allocate.
-    uint256 shares;
```
QGarchery, line 35: We should also change the dev comment above the  struct @MerlinEgalite

MerlinEgalite: cc @Rubilmax

PR #307, opened 2023-11-03
refactor(reallocate): expect final allocation

Rubilmax:
With this change, the allocator role is responsible for ordering input allocations so as to have sufficient liquidity at each supply step

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -376,16 +376,17 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
                 totalWithdrawn += withdrawnAssets;
             } else {
-                uint256 supplied = allocation.assets.zeroFloorSub(supplyAssets);
-                if (supplied == 0) continue;
+                uint256 suppliedAssets = allocation.assets.zeroFloorSub(supplyAssets);
+                if (suppliedAssets == 0) continue;
 
                 uint256 supplyCap = config[id].cap;
                 if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
 
-                if (supplyAssets + supplied > supplyCap) revert ErrorsLib.SupplyCapExceeded(id);
+                if (supplyAssets + suppliedAssets > supplyCap) revert ErrorsLib.SupplyCapExceeded(id);
 
-                (uint256 suppliedAssets, uint256 suppliedShares) =
-                    MORPHO.supply(allocation.marketParams, supplied, 0, address(this), hex"");
+                // The market's loan asset is guaranteed to be the vault's asset because it has a non-zero supply cap.
+                (, uint256 suppliedShares) =
+                    MORPHO.supply(allocation.marketParams, suppliedAssets, 0, address(this), hex"");
 
                 emit EventsLib.ReallocateSupply(_msgSender(), id, suppliedAssets, suppliedShares);
 
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -342,8 +342,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @notice Reallocates the vault's liquidity so as to reach a given allocation of assets on each given market.
-    /// @dev The allocator can withdraw from any market, even if it's not in the withdraw queue, as long as the loan
+    /// @notice The allocator can withdraw from any market, even if it's not in the withdraw queue, as long as the loan
     /// token of the market is the same as the vault's asset.
+    /// @dev The behavior of the reallocation can be altered by state changes, including:
+    /// - Deposits on the vault that supplies to markets that are expected to be supplied to during reallocation.
+    /// - Withdrawals from the vault that withdraws from markets that are expected to be withdrawn from during
+    /// reallocation.
+    /// - Donations to the vault on markets that are expected to be supplied to during reallocation.
+    /// - Withdrawals from markets that are expected to be withdrawn from during reallocation.
+    /// @dev Any additional liquidity withdrawn during reallocation will be kept idle.
     function reallocate(MarketAllocation[] calldata allocations) external onlyAllocatorRole {
         uint256 totalSupplied;
         uint256 totalWithdrawn;
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -344,49 +344,47 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// `supplied`).
     /// @dev The allocator can withdraw from any market, even if it's not in the withdraw queue, as long as the loan
     /// token of the market is the same as the vault's asset.
-    function reallocate(MarketAllocation[] calldata withdrawn, MarketAllocation[] calldata supplied)
-        external
-        onlyAllocatorRole
-    {
-        uint256 totalWithdrawn;
-        for (uint256 i; i < withdrawn.length; ++i) {
-            MarketAllocation memory allocation = withdrawn[i];
-            Id id = allocation.marketParams.id();
-
-            if (allocation.marketParams.loanToken != asset()) {
-                revert ErrorsLib.InconsistentAsset(id);
-            }
-
-            // Guarantees that unknown frontrunning donations can be withdrawn, in order to disable a market.
-            uint256 shares;
-            if (allocation.assets == type(uint256).max) {
-                shares = MORPHO.supplyShares(id, address(this));
-
-                allocation.assets = 0;
-            }
-
-            (uint256 withdrawnAssets,) =
-                MORPHO.withdraw(allocation.marketParams, allocation.assets, shares, address(this), address(this));
-
-            totalWithdrawn += withdrawnAssets;
-        }
-
+    function reallocate(MarketAllocation[] calldata totalAllocation) external onlyAllocatorRole {
         uint256 totalSupplied;
-        for (uint256 i; i < supplied.length; ++i) {
-            MarketAllocation memory allocation = supplied[i];
+        uint256 totalWithdrawn;
+        for (uint256 i; i < totalAllocation.length; ++i) {
+            MarketAllocation memory allocation = totalAllocation[i];
             Id id = allocation.marketParams.id();
-            uint256 supplyCap = config[id].cap;
 
-            if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
+            if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
 
-            (uint256 suppliedAssets,) =
-                MORPHO.supply(allocation.marketParams, allocation.assets, 0, address(this), hex"");
+            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) =
+                MORPHO.expectedMarketBalances(allocation.marketParams);
 
-            if (_supplyBalance(allocation.marketParams) > supplyCap) {
-                revert ErrorsLib.SupplyCapExceeded(id);
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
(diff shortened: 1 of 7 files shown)


Rubilmax:
For reference, the gas diff is:
image

TLDR ; surprisingly, this version costs less. It may be explained by the fact that hardhat tests previously were withdrawing all then supplying some and now are only purely "reallocating" so they do less interactions with Morpho on avg

So this result is biased

MerlinEgalite:
I'm not sure I prefer this version and what is the most intuitive

src/MetaMorpho.sol
```diff
@@ -344,49 +344,47 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// `supplied`).
     /// @dev The allocator can withdraw from any market, even if it's not in the withdraw queue, as long as the loan
     /// token of the market is the same as the vault's asset.
-    function reallocate(MarketAllocation[] calldata withdrawn, MarketAllocation[] calldata supplied)
-        external
-        onlyAllocatorRole
-    {
+    function reallocate(MarketAllocation[] calldata totalAllocation) external onlyAllocatorRole {
```
Jean-Grimal, line 347:
```suggestion
    function reallocate(MarketAllocation[] calldata allocation) external onlyAllocatorRole {
```
Couldn't it just be named allocation ?

Jean-Grimal (commented):
I actually like this implementation

src/MetaMorpho.sol
Rubilmax: How do name each element of the array then?

Jean-Grimal: Oh yes mb

Jean-Grimal (approved):

MerlinEgalite (approved):

QGarchery:
I think I lack some context on this PR.

> It is still possible to front-run reallocate and donate to the first withdrawn market up to the expected assets + idle so that the vault lacks liquidity. But it requires even more donations from the attacker

Why would it lack liquidity, isn't it just adding more to the idle liquidity ?

> it solves any front-running issues though

It's contradictory with the description "it is still possible to front-run reallocate" no ?

MerlinEgalite (approved):

QGarchery:
> Actually you're right it's just adding to the idle liquidity

So it does solve any frontrunning issue, right ? Because the only thing the attacker can do is to add to the idle liquidity (or to any market, but that is always possible)

Rubilmax:
> So it does solve any frontrunning issue, right ? Because the only thing the attacker can do is to add to the idle liquidity (or to any market, but that is always possible)

In the end I agree! Removed the PR's description, waiting for @StErMi 's approval

src/MetaMorpho.sol
```diff
@@ -345,6 +345,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// `supplied`).
     /// @dev The allocator can withdraw from any market, even if it's not in the withdraw queue, as long as the loan
     /// token of the market is the same as the vault's asset.
+
```
MerlinEgalite, line 348: to remove i think

MerlinEgalite (changes requested):

MerlinEgalite (approved):

src/MetaMorpho.sol
```diff
@@ -345,51 +345,52 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-                revert ErrorsLib.SupplyCapExceeded(id);
-            }
+                (uint256 suppliedAssets, uint256 suppliedShares) =
+                    MORPHO.supply(allocation.marketParams, supplied, 0, address(this), hex"");
 
-            totalSupplied += suppliedAssets;
+                emit EventsLib.ReallocateSupply(_msgSender(), id, suppliedAssets, suppliedShares);
 
-            emit EventsLib.ReallocateSupply(_msgSender(), id, suppliedAssets, suppliedShares);
+                totalSupplied += suppliedAssets;
```
QGarchery, line 392: Curious to know why you switched those 2 lines :eyes:

src/MetaMorpho.sol
```diff
@@ -345,51 +345,52 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+                uint256 supplied = allocation.assets.zeroFloorSub(supplyAssets);
+                if (supplied == 0) continue;
 
-            if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
+                uint256 supplyCap = config[id].cap;
+                if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
 
-            (uint256 suppliedAssets, uint256 suppliedShares) =
-                MORPHO.supply(allocation.marketParams, allocation.assets, 0, address(this), hex"");
+                if (supplyAssets + supplied > supplyCap) revert ErrorsLib.SupplyCapExceeded(id);
```
QGarchery, line 385: It's minor but this does not compute exactly the same thing as if the check was done after the `supply` call: the variable `supplyAssets` is computed using old values of the total supply and the total supply shares

src/MetaMorpho.sol
```diff
@@ -345,51 +345,52 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+    function reallocate(MarketAllocation[] calldata allocations) external onlyAllocatorRole {
+        uint256 totalSupplied;
         uint256 totalWithdrawn;
-        for (uint256 i; i < withdrawn.length; ++i) {
-            MarketAllocation memory allocation = withdrawn[i];
+        for (uint256 i; i < allocations.length; ++i) {
+            MarketAllocation memory allocation = allocations[i];
             Id id = allocation.marketParams.id();
 
             if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
```
QGarchery, line 355: This check is now done for both the supply and the withdraw case, but it is redundant in the supply case because we already check the cap (and a market with cap > 0 cannot have an inconsistent asset)

src/MetaMorpho.sol
```diff
@@ -345,51 +345,52 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// `supplied`).
```
QGarchery, line 345: We should update this comment: the allocators can now choose freely a list of supply/withdraw operations to do. I think it should just mention that this is the final allocation

QGarchery (changes requested):
The frontrunning "attacks" that it prevent does not seem very scary because the allocators could always reallocate with a margin and that would mean that the attacker donates a lot. At the same time this implementation is pretty nice so I'm not against doing those changes

src/MetaMorpho.sol
Rubilmax: This is detail, but I feel like:
- emitting the event right after the CALL is cleaner than after having increased the memory variable
- it allows to free the stack from `suppliedShares`

src/MetaMorpho.sol
Rubilmax: Updated, is it ok?

src/MetaMorpho.sol
Rubilmax: Why would they be "old"? Old compared to what? These values are not expected to change in this execution context right?

MerlinEgalite:
Since Morpho Blue has been updated the PR is broken now @Rubilmax (the naming of getters have changed)

src/MetaMorpho.sol
```diff
@@ -341,55 +341,56 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-            if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
+                totalWithdrawn += withdrawnAssets;
+            } else {
+                uint256 supplied = allocation.assets.zeroFloorSub(supplyAssets);
+                if (supplied == 0) continue;
 
-            (uint256 suppliedAssets, uint256 suppliedShares) =
-                MORPHO.supply(allocation.marketParams, allocation.assets, 0, address(this), hex"");
+                uint256 supplyCap = config[id].cap;
+                if (supplyCap == 0) revert ErrorsLib.UnauthorizedMarket(id);
```
StErMi, line 383: I know that these are details, but I think that it's worth growing these habits. 

The `supplyCap == 0` means two things:
1) explicit: the curator has disabled the market, and you cannot supply to it (or the market has never been enabled)
2) implicit the market that you are trying to supply to have not the same asset as the Metamorpho vault

The first one is easy to see, the second one not. A dev comment would be nice

MerlinEgalite: Are you suggesting we should we should revert differently depending on the case?

src/MetaMorpho.sol
```diff
@@ -345,51 +345,52 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
             Id id = allocation.marketParams.id();
 
             if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
 
-            // Guarantees that unknown frontrunning donations can be withdrawn, in order to disable a market.
-            uint256 shares;
-            if (allocation.assets == type(uint256).max) {
-                shares = MORPHO.supplyShares(id, address(this));
+            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) =
+                MORPHO.expectedMarketBalances(allocation.marketParams);
```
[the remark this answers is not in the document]
Rubilmax: done in #313

MerlinEgalite (approved):

src/MetaMorpho.sol
Rubilmax: I added a comment as suggested

StErMi: > Are you suggesting that we should revert differently depending on the case?

this depends mostly on you and what you want to monitor. the function is authed so I assume that the person should be trusted. 

it would be needed if you want to monitor an authed user that is trying to supply to a not authorized (asset different from the vault) market. but you have to be very stupid to do that knowing that it would anyway revert

MerlinEgalite (approved):

MerlinEgalite:
@Jean-Grimal @QGarchery can you check this one please?

Jean-Grimal:
> @Jean-Grimal @QGarchery can you check this one please?

LGTM. This implementation feels more natural no me. It really matches the function's name

src/MetaMorpho.sol
```diff
@@ -650,7 +658,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @dev Returns the vault's balance the market defined by `marketParams`.
     function _supplyBalance(MarketParams memory marketParams) internal view returns (uint256) {
-        return MORPHO.expectedSupplyBalance(marketParams, address(this));
+        return MORPHO.expectedSupplyAssets(marketParams, address(this));
```
StErMi, line 661: Nitpick: the function name should match the underlying function's name called. This also solves the name weirdness that right now could seem to imply that you are going to **supply** (as an action) instead of retrieving the value.

StErMi:
Am I wrong, or before, you were calling the `morpho.accrueInterest` during the reallocate iteration? 
Why have you removed it?

MerlinEgalite:
> Am I wrong, or before, you were calling the `morpho.accrueInterest` during the reallocate iteration? Why have you removed it?

This is a PR on top in fact: https://github.com/morpho-org/metamorpho/pull/313

We separated the changes

StErMi:
> > Am I wrong, or before, you were calling the `morpho.accrueInterest` during the reallocate iteration? Why have you removed it?
> 
> This is a PR on top in fact: #313
> 
> We separated the changes

Please let me know when you have finished all the work and which is the final PR I need to review because it seems that #313 is doing quite many changes

QGarchery (approved):

PR #310, opened 2023-11-07
fix(fee): update last total assets before transferring underlying

### LINKED ISSUE #220: Add tests `totalAssets == lastTotalAssets` after each interaction that update `lastTotalAssets`

### LINKED ISSUE #300: M-02 Metamorpho Vaults With ERC-777 Tokens Can Be Drained in Favor of the feeRecipient
Screenshot 2023-11-06 at 18 41 08

Rubilmax:
- Fixes #300 
- Fixes #220

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -499,6 +499,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 newTotalAssets = _accrueFee();
 
         shares = _convertToSharesWithTotals(assets, totalSupply(), newTotalAssets, Math.Rounding.Floor);
+
+        // `newTotalAssets + assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets + assets);
+
         _deposit(_msgSender(), receiver, assets, shares);
     }
 
@@ -507,6 +511,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 newTotalAssets = _accrueFee();
 
         assets = _convertToAssetsWithTotals(shares, totalSupply(), newTotalAssets, Math.Rounding.Ceil);
+
+        // `newTotalAssets + assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets + assets);
+
         _deposit(_msgSender(), receiver, assets, shares);
     }
 
@@ -521,6 +529,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         // Do not call expensive `maxWithdraw` and optimistically withdraw assets.
 
         shares = _convertToSharesWithTotals(assets, totalSupply(), newTotalAssets, Math.Rounding.Ceil);
+
+        // `newTotalAssets - assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets.zeroFloorSub(assets));
+
         _withdraw(_msgSender(), receiver, owner, assets, shares);
     }
 
@@ -535,6 +547,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         // Do not call expensive `maxRedeem` and optimistically redeem shares.
 
         assets = _convertToAssetsWithTotals(shares, totalSupply(), newTotalAssets, Math.Rounding.Floor);
+
+        // `newTotalAssets - assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets.zeroFloorSub(assets));
+
         _withdraw(_msgSender(), receiver, owner, assets, shares);
     }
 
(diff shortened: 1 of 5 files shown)


MerlinEgalite (commented):
Too bad we lose the clean accounting but it's for the better. Perhaps should we comment near `lastTotalAssets`, that this value can be differ differ from the theoretical value?

src/MetaMorpho.sol
```diff
@@ -520,15 +521,23 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+
         _deposit(_msgSender(), receiver, assets, shares);
     }
 
     /// @inheritdoc IERC4626
     function mint(uint256 shares, address receiver) public override returns (uint256 assets) {
         uint256 newTotalAssets = _accrueFee();
 
+        // `newTotalAssets + assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets + assets);
```
Rubilmax, line 537:
```suggestion
        lastTotalAssets = newTotalAssets;
```

src/MetaMorpho.sol
```diff
@@ -520,15 +521,23 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     function deposit(uint256 assets, address receiver) public override returns (uint256 shares) {
         uint256 newTotalAssets = _accrueFee();
 
+        // `newTotalAssets + assets` may be a little off from `totalAssets()`.
+        _updateLastTotalAssets(newTotalAssets + assets);
```
Rubilmax, line 525:
```suggestion
        lastTotalAssets = newTotalAssets;
```

Rubilmax: 1. Otherwise we emit the event twice
2. At this stage, only newTotalAssets are on the vault

src/MetaMorpho.sol
```diff
@@ -520,15 +521,25 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+
         _deposit(_msgSender(), receiver, assets, shares);
     }
 
     /// @inheritdoc IERC4626
     function mint(uint256 shares, address receiver) public override returns (uint256 assets) {
         uint256 newTotalAssets = _accrueFee();
 
+        // Update `lastTotalAssets` a first time to prevent reentrancy attacks.
+        // It is re-updated in `_deposit`.
```
Rubilmax, line 538:
```suggestion
        // Update `lastTotalAssets` to avoid an inconsistent state in a re-entrant context. It is updated again in `_deposit`.
```

src/MetaMorpho.sol
```diff
@@ -520,15 +521,25 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     function deposit(uint256 assets, address receiver) public override returns (uint256 shares) {
         uint256 newTotalAssets = _accrueFee();
 
+        // Update `lastTotalAssets` a first time to prevent reentrancy attacks.
+        // It is re-updated in `_deposit`.
```
Rubilmax, line 525:
```suggestion
        // Update `lastTotalAssets` to avoid an inconsistent state in a re-entrant context. It is updated again in `_deposit`.
```

src/MetaMorpho.sol
```diff
@@ -628,7 +647,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
         _supplyMorpho(assets);
 
-        // `newTotalAssets + assets` cannot be used as input because of rounding errors so we must use `totalAssets`.
+        // Update `lastTotalAssets` again to avoid inconsistent state due to a reentrancy for instance.
         _updateLastTotalAssets(totalAssets());
```
Rubilmax, line 651:
```suggestion
        // `lastTotalAssets + assets` may be a little off from `totalAssets()`.
        _updateLastTotalAssets(lastTotalAssets + assets);
```

MerlinEgalite: It's not symmetric deposit/mint then and introduce small rounding. Are we sure we want to do that?

Rubilmax: > It's not symmetric deposit/mint

I don't get the symmetry you're talking about ; to me there's no symmetry nowhere

> introduce small rounding

Yes, as in withdraw/redeem
If we don't want this rounding in the end, we should revert the change in withdraw/redeem (and keep the previous comments about why we're doing so)

MerlinEgalite: In `mint`/`deposit`/`withdraw`/`redeem` the `lastTotalAssets` is updated using `totalAssets()` ultimately which would not be the case anymore with your proposition for `mint`/`deposit` only

Rubilmax: That is incorrect: with the current changes to withdraw and redeem, lastTotalAssets is no longer calculated using totalAssets()

Hence why I'm suggesting to add assets instead of calling totalAssets() in deposit/mint: to be consistent
But we can be consistent the other way around by reverting the changes to withdraw & redeem

MerlinEgalite: This is the current state of the code:

https://github.com/morpho-org/metamorpho/blob/4220004a6a38caacaff31a8af0e67a1ccdef5842/src/MetaMorpho.sol#L651

Rubilmax: Yes and I'm highlighting that there's no symmetry with withdraw/redeem: https://github.com/morpho-org/metamorpho/blob/4220004a6a38caacaff31a8af0e67a1ccdef5842/src/MetaMorpho.sol#L568-L569

PR #312, opened 2023-11-07
Fix dead IRM market

Rubilmax:
Replaces #281 

Does not include the feature request from @QGarchery to separate roles: with this PR, the curator can allow a market to be freely removed by any allocator after some timelock

I'm writing this feature request on top


src/MetaMorpho.sol
```diff
@@ -284,6 +285,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
+    function disableMarket(Id id) external onlyCuratorRole {
```
Jean-Grimal, line 288:
```suggestion
    function prepareMarketdisabling(Id id) external onlyCuratorRole {
```
`prepareMarketdisabling ` is ugly but I'm sure we can find better names for this. 

The fact is that function is just the first step for disabling a market.

test/forge/MarketTest.sol
Jean-Grimal, line 1: `disableMarket`should be tested as well

test/metamorpho_tests.tree
Jean-Grimal, line 1: I'll add the BTT of `disableMarket`

Jean-Grimal (changes requested):

Jean-Grimal:
Guardian should be able to revoke market disabling.

src/MetaMorpho.sol
MerlinEgalite, line 288: missing nastpecs

MerlinEgalite: `submitDisableMarket` let's keep the classic name

src/MetaMorpho.sol
```diff
@@ -687,11 +700,11 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                 revert ErrorsLib.MaxQueueLengthExceeded();
             }
 
-            // Safe "unchecked" cast because withdrawQueue.length <= MAX_QUEUE_LENGTH.
-            marketConfig.withdrawRank = uint64(withdrawQueue.length);
+            marketConfig.enabled = true;
```
MerlinEgalite, line 703: we're mixing changes no?

src/libraries/PendingLib.sol
```diff
@@ -1,18 +1,27 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.0;
 
+struct MarketConfig {
+    /// @notice The maximum amount of assets that can be allocated to the market.
+    uint192 cap;
+    /// @notice Whether the market is in the withdraw queue.
+    bool enabled;
+    /// @notice The timestamp at which the market can be instantly removed from the withdraw queue.
+    uint48 removableAt;
```
MerlinEgalite, line 10:
```suggestion
    uint56 removableAt;
```

and we update the rest no?

src/MetaMorpho.sol
```diff
@@ -687,11 +700,11 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
         }
 
         marketConfig.cap = supplyCap;
+        marketConfig.removableAt = 0;
```
MerlinEgalite, line 707: so setting a cap reset the action of disabling a market

src/MetaMorpho.sol
```diff
@@ -326,20 +336,23 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-                if (MORPHO.supplyShares(id, address(this)) != 0 || config[id].cap != 0) {
-                    revert ErrorsLib.InvalidMarketRemoval(id);
+                if (config[id].cap != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroCap(id);
+
+                if (config[id].removableAt != 0) {
+                    if (block.timestamp < config[id].removableAt) {
+                        revert ErrorsLib.InvalidMarketRemovalTimelockNotElapsed(id);
+                    }
+                } else if (MORPHO.supplyShares(id, address(this)) != 0) {
+                    revert ErrorsLib.InvalidMarketRemovalNonZeroSupply(id);
```
MerlinEgalite, line 352:
```suggestion
                if (config[id].removableAt != 0 && block.timestamp < config[id].removableAt) {
                        revert ErrorsLib.InvalidMarketRemovalTimelockNotElapsed(id);
                    }
                } else if (MORPHO.supplyShares(id, address(this)) != 0) {
                    revert ErrorsLib.InvalidMarketRemovalNonZeroSupply(id);
```

src/MetaMorpho.sol
```diff
@@ -284,6 +285,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
+    function disableMarket(Id id) external onlyCuratorRole {
+        if (config[id].removableAt != 0) revert ErrorsLib.AlreadySet();
+        if (config[id].cap == 0) revert ErrorsLib.MarketNotEnabled();
+
+        config[id].disable(timelock);
+
+        emit EventsLib.DisableMarket(_msgSender(), id);
```
MerlinEgalite, line 294:
```suggestion
        emit EventsLib.SubmitPendingDisableMarket(_msgSender(), id);
```

MerlinEgalite (changes requested):
The `guardian` should be able to cancel disabling a market maybe

src/MetaMorpho.sol
Rubilmax: No this PR is also based on #282

src/MetaMorpho.sol
Rubilmax: The suggested behavior is not the same and is faulty: if some market is removable (the timelock timestamp has passed), then the suggested code requires the removable market to have 0 supply shares

src/MetaMorpho.sol
Rubilmax: Yes otherwise an allocator can freely remove a previously disabled market which was re-enabled

src/MetaMorpho.sol
```diff
@@ -284,6 +285,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
+    function disableMarket(Id id) external onlyCuratorRole {
+        if (config[id].removableAt != 0) revert ErrorsLib.AlreadySet();
+        if (config[id].cap == 0) revert ErrorsLib.MarketNotEnabled();
```
Rubilmax, line 290: This last line has non-trivial implications: if the curator by mistake reacts fast and sets the cap of a market to zero, they can't disable it. They need to first set a non-zero cap, wait for the timelock, accept the cap then submit the market's removal and wait for the timelock again

Rubilmax: We could:
1. acknowledge this
2. remove this line and not set `cap = 0` upon market removal and expect the curator to do this step on their own during the timelock
3. not require `cap == 0` but still enforce it for protection

src/MetaMorpho.sol
MerlinEgalite: You're right

src/MetaMorpho.sol
MerlinEgalite: it's more a not to my self. I wonder if we should document it or leave as it is

src/MetaMorpho.sol
MerlinEgalite: I'm in favor of option 3.

The current implementation can be very problematic I think

src/MetaMorpho.sol
Rubilmax: I actually moved it to where we set `enabled = false`
So it is always possible to have a market with a non-zero cap and a timelock for it to be removable

src/MetaMorpho.sol
Rubilmax: Actually option 3 is weird because if the guardian revokes the removal of the market, the cap would be left with 0 (we can't recover the previous cap)
I'd argue to expect the curator to bundle the cap set to 0 and the market's removal instead, so that each function is independent

Rubilmax: Arg but if we remove it we can submit a non-enabled market for removal

README.md
```diff
@@ -54,28 +56,37 @@ It can:
...
 It can:
+
 - Do what the allocators can do.
-- [Timelocked] Enable or disable a market by setting a cap to a specific market.
-    - The cap must be set to 0 to disable the market.
-	- Disabling a market can then only be done if the vault has no liquidity supplied on the market.
+- Decrease the supply cap of any market.
+  - To softly remove a market, it is expected from the allocator role to reallocate the supplied liquidity to another enabled market first.
+- [Timelocked] Increase the supply cap of any market.
+- [Timelocked] Submits the forced removal of a market.
```
MerlinEgalite, line 64:
```suggestion
- [Timelocked] Submit the forced removal of a market.
```

README.md
```diff
@@ -54,28 +56,37 @@ It can:
...
+  - This action is typically designed to force the removal of a market that keeps reverting thus locks the vault.
```
MerlinEgalite, line 65:
```suggestion
  - This action is typically designed to force the removal of a market that keeps reverting thus locking the vault.
```

README.md
```diff
@@ -54,28 +56,37 @@ It can:
...
+  - After the timelock has elapsed, the allocator role is free to remove the market from the withdraw queue. The funds supplied to this market will be lost.
+  - If the market ever functions again, the allocator role can withdraw the funds that were previously lost.
```
MerlinEgalite, line 67: In the case it has not been removed, right?

src/MetaMorpho.sol
```diff
@@ -446,6 +449,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         emit EventsLib.RevokePendingCap(_msgSender(), id);
     }
 
+    function revokePendingMarketRemoval(Id id) external onlyGuardian {
```
MerlinEgalite, line 451: Missing natspecs

src/libraries/PendingLib.sol
```diff
@@ -34,22 +34,21 @@ library PendingLib {
...
         pending.value = newValue;
         // Safe "unchecked" cast because timelock <= MAX_TIMELOCK.
-        pending.validAt = uint48(block.timestamp) + uint48(timelock);
+        pending.validAt = uint56(block.timestamp + timelock);
     }
 
     /// @dev Marks the market associated to the given config as removable starting `timelock` seconds in the future.
     /// @dev Assumes `timelock` <= `MAX_TIMELOCK`.
     function disable(MarketConfig storage config, uint256 timelock) internal {
-        config.cap = 0;
```
MerlinEgalite: I would keep it

src/MetaMorpho.sol
```diff
@@ -285,13 +285,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
-    function disableMarket(Id id) external onlyCuratorRole {
+    /// @notice Submits a forced market removal from the vault, losing all funds supplied to the market.
+    /// @dev To remove it from the withdraw queue after the timelock, the market also needs to have a zero cap.
+    /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
+    function submitMarketRemoval(Id id) external onlyCuratorRole {
         if (config[id].removableAt != 0) revert ErrorsLib.AlreadySet();
-        if (config[id].cap == 0) revert ErrorsLib.MarketNotEnabled();
```
MerlinEgalite: why don't we do the opposite ie we require that the cap is 0?

src/MetaMorpho.sol
```diff
@@ -353,6 +355,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                 }
 
                 config[id].enabled = false;
```
MerlinEgalite, line 357:
```suggestion
                delete config[id];
```

MerlinEgalite (changes requested):

README.md
Rubilmax: No, the allocator can always withdraw liquidity from non-enabled markets

src/libraries/PendingLib.sol
Rubilmax: Can we continue the discussion here: https://github.com/morpho-org/metamorpho/pull/312#discussion_r1386189685 and can you argue?

src/MetaMorpho.sol
Rubilmax: @MerlinEgalite :
> why don't we do the opposite ie we require that the cap is 0?

This gave me a better idea: require the `enabled` flag, see latest changes

src/MetaMorpho.sol
Rubilmax: discussion merged with: https://github.com/morpho-org/metamorpho/pull/312#discussion_r1386302217

src/MetaMorpho.sol
```diff
@@ -284,6 +285,18 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
     }
 
+    /// @notice Submits a forced market removal from the vault, losing all funds supplied to the market.
```
Jean-Grimal, line 288:
```suggestion
    /// @notice Submits a forced market removal from the vault, eventually losing all funds supplied to the market.
```
Because if the market stops reverting, the funds might be recovered

src/MetaMorpho.sol
```diff
@@ -284,6 +285,18 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+    /// @dev To remove it from the withdraw queue after the timelock, the market also needs to have a zero cap.
```
Jean-Grimal, line 289:
```suggestion
    /// @dev To remove it from the withdraw queue after the timelock.
```
This function already sets the cap of the market to zero.

src/MetaMorpho.sol
```diff
@@ -284,6 +285,18 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+    /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
+    function submitMarketRemoval(Id id) external onlyCuratorRole {
```
Jean-Grimal, line 291:
```suggestion
    function submitMarketForceRemoval(Id id) external onlyCuratorRole {
```

src/MetaMorpho.sol
```diff
@@ -433,6 +451,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         emit EventsLib.RevokePendingCap(_msgSender(), id);
     }
 
+    /// @notice Revokes the pending removal of the market defined by `id`.
+    function revokePendingMarketRemoval(Id id) external onlyGuardian {
```
Jean-Grimal, line 455:
```suggestion
    function revokePendingMarketForceRemoval(Id id) external onlyGuardian {
```

src/libraries/ErrorsLib.sol
```diff
@@ -39,8 +39,14 @@ library ErrorsLib {
...
-    /// @notice Thrown when market `id` is missing in the new withdraw queue to set.
-    error InvalidMarketRemoval(Id id);
+    /// @notice Thrown when market `id` is missing in the updated withdraw queue and the market has a non-zero cap set.
+    error InvalidMarketRemovalNonZeroCap(Id id);
+
+    /// @notice Thrown when market `id` is missing in the updated withdraw queue and the market has a non-zero supply.
+    error InvalidMarketRemovalNonZeroSupply(Id id);
+
+    /// @notice Thrown when market `id` is missing in the updated withdraw queue and the market is not yet disabled.
+    error InvalidMarketRemovalTimelockNotElapsed(Id id);
```
Jean-Grimal, line 49:
```suggestion
    error InvalidMarketForceRemovalTimelockNotElapsed(Id id);
```

Jean-Grimal (changes requested):

src/MetaMorpho.sol
Rubilmax: I think it's unnecessary and verbose, wdyt @morpho-org/onchain-team ?

I understand though that a "market removal" can be done in 2 ways:
1. via this timelocked function then with update withdraw queue
2. via cap set to zero, reallocation and update withdraw queue

And it may be confusing

MerlinEgalite: I think it's too verbose as well and I would keep the current version

MerlinEgalite (approved):

Jean-Grimal (approved):

MerlinEgalite (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):

MerlinEgalite (approved):

QGarchery (changes requested):
A few things to change but otherwise I like the idea

src/MetaMorpho.sol
```diff
@@ -513,7 +514,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @notice Accepts the pending cap of the market defined by `id`.
     function acceptCap(Id id) external afterTimelock(pendingCap[id].validAt) {
-        _setCap(id, pendingCap[id].value);
+        // Safe "unchecked" cast because pendingCap <= type(uint184).max.
```
QGarchery, line 517: We could go for uint184 everywhere so that: we are only using `toUint184`, and we don't need to verify this property to make sure that this cast is safe. According to my tests it does not even cost more gas

QGarchery (approved):

MerlinEgalite (approved):

PR #313, opened 2023-11-07
Perf/accrue interest

Rubilmax:


src/MetaMorpho.sol
```diff
@@ -649,9 +645,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         return IMorphoMarketParams(address(MORPHO)).idToMarketParams(id);
     }
 
-    /// @dev Returns the vault's balance the market defined by `marketParams`.
-    function _supplyBalance(MarketParams memory marketParams) internal view returns (uint256) {
-        return MORPHO.expectedSupplyAssets(marketParams, address(this));
+    function _accruedSupplyBalance(MarketParams memory marketParams, Id id)
```
MerlinEgalite, line 648: Lack of natspecs

src/MetaMorpho.sol
```diff
@@ -794,30 +820,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-    function _suppliable(MarketParams memory marketParams, Id id) internal view returns (uint256) {
-        uint256 supplyCap = config[id].cap;
-        if (supplyCap == 0) return 0;
-
-        return supplyCap.zeroFloorSub(_supplyBalance(marketParams));
-    }
-
-    /// @dev Returns the withdrawable amount of assets from the market defined by `marketParams`.
-    /// @dev Assumes that the inputs `marketParams` and `id` match.
-    function _withdrawable(MarketParams memory marketParams, Id id) internal view returns (uint256) {
```
MerlinEgalite, line 808: I would keep this function, we're duplicating code now

src/MetaMorpho.sol
```diff
@@ -777,11 +791,23 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             Id id = withdrawQueue[i];
             MarketParams memory marketParams = _marketParams(id);
 
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
```
MerlinEgalite, line 794: before the code was much simpler

MerlinEgalite (changes requested):
I get why we want to avoid performing twice the same computation and calls but I don't get why we would like to duplicate the code at the same time

src/MetaMorpho.sol
Rubilmax: We're not duplicating code: `_withdrawable` was used once in `_simulateWithdrawMorpho` virtually accruing interest and once in `_withdrawMorpho` which now accrues interest on Morpho directly

src/MetaMorpho.sol
Rubilmax: You mean because we were internalizing half of the code into `_suppliable`?

src/MetaMorpho.sol
Rubilmax: I factorized it, wdyt?

MerlinEgalite: it's better to me thx

MerlinEgalite (approved):

Jean-Grimal (approved):

src/MetaMorpho.sol
```diff
@@ -649,9 +645,18 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+    /// @dev Accrues interest on Morpho Blue and returns the vault's assets & corresponding shares supplied on the
+    /// market defined by `marketParams`, as well as the market's state.
+    /// @dev Assumes that the inputs `marketParams` and `id` match.
+    function _accruedSupplyBalance(MarketParams memory marketParams, Id id)
+        internal
+        returns (uint256 assets, uint256 shares, Market memory market)
+    {
+        MORPHO.accrueInterest(marketParams);
+
+        market = MORPHO.market(id);
```
StErMi, line 657: Is it beneficial to have a utility function that, given a `marketID` and `userAddress` returns both the market data and user data?

It would avoid you to perform multiple external calls. But I don't want you to bloat Morpho Blue just because of a one-time thing.

It's anyway something that you should consider evaluating: what utility getters are needed for you (metamorpho, bundlers, URD, and so on) and integrators?

MerlinEgalite: I think one can leverage the `MorphoStorageLib` to extract more efficiently data but it would complexify a bit the development. I think the current version is fine tbh

PR #314, opened 2023-11-08
Enable tracking the accrued interest

QGarchery:
Motivated by a remark of @tomrpl, thanks !

Before this PR, the events `AccrueFee(uint256 feeShares)` and `UpdateLastTotalAssets(uint256 newTotalAssets)` were emitted at each interaction. Those are not enough to compute the actual yield generated by the vault because:
- `feeShares` could be 0, if the vault does not take any performance fee we can't recompute the whole interest generated
- `newTotalAssets` takes into account the funds added by the current interaction. Those funds are not part of the interest generated and are difficult to subtract (it depends on each interaction)

To fix this, the event `AccrueFee` is transformed into `AccrueInterest` to add 2 fields:
1. `lastTotalAssets`, the value of the storage variable of the same name before the interaction. It is provided as a convenience to be able to reconstruct the yield by looking only at events
2. `accruedTotalAssets`, the total assets in the vault before the interaction, after having accrued the interest

With those, one can compute `(accruedTotalAssets - lastTotalAssets) / lastTotalAssets` to compute the yield generated during 2 interactions.

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -823,14 +823,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @dev Accrues the fee and mints the fee shares to the fee recipient.
-    /// @return newTotalAssets The new vault's total assets.
-    function _accrueFee() internal returns (uint256 newTotalAssets) {
+    /// @return accruedTotalAssets The vaults total assets after accruing the interest.
+    function _accrueFee() internal returns (uint256 accruedTotalAssets) {
         uint256 feeShares;
-        (feeShares, newTotalAssets) = _accruedFeeShares();
+        (feeShares, accruedTotalAssets) = _accruedFeeShares();
 
         if (feeShares != 0) _mint(feeRecipient, feeShares);
 
-        emit EventsLib.AccrueInterest(lastTotalAssets, newTotalAssets, feeShares);
+        emit EventsLib.AccrueInterest(lastTotalAssets, accruedTotalAssets, feeShares);
     }
 
     /// @dev Computes and returns the fee shares (`feeShares`) to mint and the new vault's total assets
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -828,11 +828,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         uint256 feeShares;
         (feeShares, newTotalAssets) = _accruedFeeShares();
 
-        if (feeShares != 0) {
-            _mint(feeRecipient, feeShares);
+        if (feeShares != 0) _mint(feeRecipient, feeShares);
 
-            emit EventsLib.AccrueFee(feeShares);
-        }
+        emit EventsLib.AccrueInterest(lastTotalAssets, newTotalAssets, feeShares);
     }
 
     /// @dev Computes and returns the fee shares (`feeShares`) to mint and the new vault's total assets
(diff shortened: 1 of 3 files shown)


src/MetaMorpho.sol
```diff
@@ -823,16 +823,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @dev Accrues the fee and mints the fee shares to the fee recipient.
-    /// @return newTotalAssets The new vault's total assets.
-    function _accrueFee() internal returns (uint256 newTotalAssets) {
+    /// @return accruedTotalAssets The vaults total assets after accruing the interest.
```
Rubilmax, line 826: I actually find the naming `accruedTotalAssets` to be confusing, because I would expect it to correspond to `totalAssets() - lastTotalAssets` and not only `totalAssets()` as it is now

Am I the only one?

Rubilmax (commented):
Ouf, missed this one thanks

src/MetaMorpho.sol
QGarchery: I see what you mean but with `total` in the name it should lead in the right direction no ? 

Do you have a suggestion ?

Rubilmax: `Total` doesn't lead me to the right direction still, because I would think that it corresponds to the "accrued difference of total assets"

But I'm the only one then I guess I'm too biased

I preferred the naming "newTotalAssets" which I agree can be confusing from a macro perspective where you have 2 "new total assets" values for each interaction but in the context of the event it may be fine because you only have a single possible "new total assets" after having accrued interest

QGarchery: Having twice the same name is confusing to me. What do you think about `totalAssetsAfterAccrual` ?

QGarchery: Or `afterAccrualTotalAssets`, to stay consistent with `lastTotalAssets`and `newTotalAssets` that prefix `TotalAssets`

Rubilmax: It works, though it's verbose

MerlinEgalite: `updatedTotalAssets`?

julien-devatom:
`lastTotalAssets` can be recovered from the previous event, right?

Rubilmax:
> `lastTotalAssets` can be recovered from the previous event, right?

No because the `totalAssets` emitted in this event does not include the eventual deposit/withdrawal that happens after

src/MetaMorpho.sol
QGarchery: > Where is the name used twice?

In the event `UpdateLastTotalAssets`.

> `updatedTotalAssets`?

This is also confusing, because it is not the value that is used to update total assets stored (and there is the `UpdateLastTotalAssets` event that we could think is linked to it)

QGarchery: So instead I just changed the event `UpdateLastTotalAssets` to use the name `updatedTotalAssets`, which sounds obvious now :)

QGarchery:
> > `lastTotalAssets` can be recovered from the previous event, right?
> 
> No because the `totalAssets` emitted in this event does not include the eventual deposit/withdrawal that happens after

It can be recovered from the last `UpdateTotalAssets` event though. So it seems like we could remove `lastTotalAssets` from the `AccrueInterest` event. How annoying would it be to compute the yield if we did @julien-devatom ?

julien-devatom:
Its ok while we can recover it from prev event

QGarchery:
> Its ok while we can recover it from prev event

Ok, I'm removing the field `lastTotalAssets` from `AccrueInterest` then

MerlinEgalite (approved):

Rubilmax (approved):

MerlinEgalite (approved):

PR #320, opened 2023-11-10
fix(import): use relative import paths

Rubilmax:


test/forge/UrdTest.sol
```diff
@@ -1,24 +1,20 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.0;
 
-import {UtilsLib} from "@morpho-blue/libraries/UtilsLib.sol";
-
-import {UrdFactory} from "@universal-rewards-distributor/UrdFactory.sol";
-import {IUniversalRewardsDistributor} from "@universal-rewards-distributor/UniversalRewardsDistributor.sol";
```
Rubilmax, line 7: I removed these for 2 reasons:
1. they use solidity 0.8.19
2. we were essentially testing the URD, but we only want to test that MM does transfer rewards (skim tokens after #278)

Jean-Grimal (approved):

MerlinEgalite (approved):

PR #325, opened 2023-11-15
Update libs

MerlinEgalite:


Jean-Grimal (approved):

Rubilmax (commented):
Ok to replace address 0 with irm but then we need to not enable irm 0 in BaseTest

MerlinEgalite:
> Ok to replace address 0 with irm but then we need to not enable irm 0 in BaseTest

why?

PR #334, opened 2023-11-28
[Certora] Dev

QGarchery:
This PR implements the base setup of the verification. The given rule is still to verify, and some functions were temporarily filtered out

### DIFF
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/setup-python@v4
 
       - name: Install certora
-        run: pip install certora-cli
+        run: pip install certora-cli-beta
 
       - name: Install solc
         run: |
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Apply munging
         run: make -C certora munged
 
-      - name: Verify rule ${{ matrix.script }}
+      - name: Verify rule ${{ matrix.conf }}
         run: certoraRun certora/confs/${{ matrix.conf }}.conf
         env:
           CERTORAKEY: ${{ secrets.CERTORAKEY }}
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -1,0 +1,41 @@
+name: Certora
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install python
+        uses: actions/setup-python@v4
+
+      - name: Install certora
+        run: pip install certora-cli
+
+      - name: Install solc
+        run: |
+          wget https://github.com/ethereum/solidity/releases/download/v0.8.21/solc-static-linux
+          chmod +x solc-static-linux
+          sudo mv solc-static-linux /usr/local/bin/solc
+
+      - name: Verify rule ${{ matrix.script }}
+        run: certoraRun certora/confs/${{ matrix.conf }}.conf
+        env:
+          CERTORAKEY: ${{ secrets.CERTORAKEY }}
+
+    strategy:
+      fail-fast: false
+
+      matrix:
+        conf:
+          - Roles
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -11,6 +11,13 @@ jobs:
   verify:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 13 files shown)


certora/applyMunging.patch
```diff
@@ -0,0 +1,125 @@
...
+ 
+ /// @title MetaMorpho
+ /// @author Morpho Labs
+@@ -45,9 +43,7 @@
+     using UtilsLib for uint256;
+     using SafeCast for uint256;
+     using SafeERC20 for IERC20;
+-    using MorphoLib for IMorpho;
+     using SharesMathLib for uint256;
+-    using MorphoBalancesLib for IMorpho;
```
MathisGD, line 104: why can you do that ?

certora/harness/MetaMorphoHarness.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity 0.8.21;
+
+import "../munged/MetaMorpho.sol";
+
+contract MetaMorphoHarness is MetaMorpho {
```
MathisGD, line 6: is not yet IMetaMorhpoHarness right ?

certora/specs/Roles.spec
```diff
@@ -0,0 +1,56 @@
...
+    && f.selector != sig:transfer(address, uint256).selector
+    && f.selector != sig:skim(address).selector
+    && f.selector != sig:acceptCap(MetaMorphoHarness.MarketParams).selector
+    && f.selector != sig:transferFrom(address, address, uint256).selector
+    && f.selector != sig:reallocate(MetaMorphoHarness.MarketAllocation[]).selector
+    && f.selector != sig:mint(uint256, address).selector
+    && f.selector != sig:withdraw(uint256, address, address).selector
+    && f.selector != sig:multicall(bytes[]).selector
+    && f.selector != sig:submitCap(MetaMorphoHarness.MarketParams, uint256).selector
+    && f.selector != sig:acceptOwnership().selector
```
MathisGD, line 33: there is no way to do f.selector == sig:... || f.selector == sig:... instead?

MathisGD, line 33: why do we have to skip this actually ?

.github/workflows/certora.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install python
+        uses: actions/setup-python@v4
+
+      - name: Install certora
+        run: pip install certora-cli-beta
```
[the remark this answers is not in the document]
QGarchery: It is less stable, but we need it here for compiling MM: otherwise it times out for very simple rules. I don't know exactly what it does, but I'll revert the changes when the fix reaches `certora-cli` (as we have done in morpho-blue)

certora/harness/MetaMorphoHarness.sol
QGarchery: There is no IMetaMorphoHarness, because we don't really need an interface for using a contract with Certora

certora/specs/Roles.spec
QGarchery: I don't think so, but it will be temporary anyway and will be removed once this rule goes through for most functions

certora/applyMunging.patch
QGarchery: Because the interface IMorpho has been extended to IMorphoHarness, which includes all the functions which used to be defined in the periphery libraries of Morpho Blue. It's like saying that Morpho Blue is a big contract with all the functions of the periphery defined inside. This work better with the Certora prover because otherwise the libary functions get copied many times, and this lead to a huge contract that is difficult to verify

certora/specs/Roles.spec
QGarchery: There is a pending discussion with the Certora team about how to handle summarized calls: they seem to always revert when their returned values is used, and I don't know what to do about it. Once I know more about that, I'll try to fix this rule

certora/harness/MetaMorphoHarness.sol
MathisGD: ah yes sorry I messed up. So why do we need this contract?

certora/applyMunging.patch
MathisGD: indeed, i messed up between IMorphoHarness and IMetamorphoHarness

certora/harness/MetaMorphoHarness.sol
QGarchery: This is not really needed in this PR, but it will be useful at some point so I thought we could start with it. For example, in #385 we need the harness

MathisGD (approved):

adhusson (approved):

PR #346, opened 2023-12-12
Fix rounding in max deposit

Jean-Grimal:
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/963a0dbd-fcf0-4882-9d07-ac6711667cd8
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/52649f43-6794-4046-888e-ebaaa7e3d0a4

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -773,7 +773,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (supplyCap == 0) continue;
 
             MarketParams memory marketParams = _marketParams(id);
-            (uint256 supplyAssets,,) = _accruedSupplyBalance(marketParams, id);
+
+            MORPHO.accrueInterest(marketParams);
+
+            Market memory market = MORPHO.market(id);
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+            uint256 supplyAssets = supplyShares.toAssetsUp(market.totalSupplyAssets, market.totalSupplyShares);
 
             uint256 toSupply = UtilsLib.min(supplyCap.zeroFloorSub(supplyAssets), assets);
 
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -646,10 +646,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyCap = config[id].cap;
             if (supplyCap == 0) continue;
 
-            MarketParams memory marketParams = _marketParams(id);
-
             uint256 supplyShares = MORPHO.supplyShares(id, address(this));
-            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(marketParams);
+            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(_marketParams(id));
 
             uint256 supplyAssets = supplyShares.toAssetsUp(totalSupplyAssets, totalSupplyShares);
 
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -649,9 +649,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyShares = MORPHO.supplyShares(id, address(this));
             (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(_marketParams(id));
 
-            uint256 supplyAssets = supplyShares.toAssetsUp(totalSupplyAssets, totalSupplyShares);
-
-            totalSuppliable += supplyCap.zeroFloorSub(supplyAssets);
+            totalSuppliable += supplyCap.zeroFloorSub(supplyShares.toAssetsUp(totalSupplyAssets, totalSupplyShares));
         }
     }
 
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -778,6 +778,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
             Market memory market = MORPHO.market(id);
             uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+            // `supplyAssets` needs to be rounded up for `toSupply` to be rounded down.
             uint256 supplyAssets = supplyShares.toAssetsUp(market.totalSupplyAssets, market.totalSupplyShares);
 
             uint256 toSupply = UtilsLib.min(supplyCap.zeroFloorSub(supplyAssets), assets);
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -646,7 +646,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyCap = config[id].cap;
             if (supplyCap == 0) continue;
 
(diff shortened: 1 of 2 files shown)


src/MetaMorpho.sol
```diff
@@ -646,7 +646,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyCap = config[id].cap;
             if (supplyCap == 0) continue;
 
-            uint256 supplyAssets = MORPHO.expectedSupplyAssets(_marketParams(id), address(this));
+            MarketParams memory marketParams = _marketParams(id);
+
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(marketParams);
```
Rubilmax, line 652:
```suggestion
            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(_marketParams(id));
```

Rubilmax: Optional but I feel it is welcome here

Rubilmax (commented):
Ideally we should have a test for it
Hopefully it shouldn't be too difficult

Rubilmax (dismissed):

test/forge/ERC4626Test.sol
```diff
@@ -355,6 +355,36 @@ contract ERC4626Test is IntegrationTest, IMorphoFlashLoanCallback {
...
+        morpho.supplyCollateral(allMarkets[0], 2 ether, BORROWER, hex"");
+        morpho.borrow(allMarkets[0], 1 ether, 0, BORROWER, BORROWER);
+        vm.stopPrank();
+
+        _forward(1_000);
+
+        loanToken.setBalance(SUPPLIER, 1 ether);
+
+        vm.prank(SUPPLIER);
+        vault.deposit(1 ether, ONBEHALF);
```
MerlinEgalite, line 383: why it that part necessary?

Rubilmax: If no interest is accrued then the share price is 1e-6 which doesn't incur a rounding upon supply, so the maxDeposit whether it is rounded up or down is fine
However, the issue arises as soon as the share price incurs a rounding

MerlinEgalite:
Shouldn't this issue be fixed as well: https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/52649f43-6794-4046-888e-ebaaa7e3d0a4?q=status%253Anew%252Cdisputed%252Crejected%252Cduplicate%252Cconfirmed%252Cacknowledged%252Cfixed ?

MerlinEgalite (dismissed):

src/MetaMorpho.sol
```diff
@@ -772,7 +773,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (supplyCap == 0) continue;
 
             MarketParams memory marketParams = _marketParams(id);
-            (uint256 supplyAssets,,) = _accruedSupplyBalance(marketParams, id);
+
+            MORPHO.accrueInterest(marketParams);
+
+            Market memory market = MORPHO.market(id);
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+            uint256 supplyAssets = supplyShares.toAssetsUp(market.totalSupplyAssets, market.totalSupplyShares);
```
MerlinEgalite, line 781: could we add a small comment here to say why we're not using `_accruedSupplyBalance`?

MerlinEgalite (approved):

adhusson (approved):
The  change in `_maxDeposit` reimplements `MorphoBalancesLib.expectedSupplyAssets` (rounding up), and the change in `_supplyMorpho` reimplements `MetaMorpho. _accruedSupplyBalance` (rounding up).

Not sure how to remember this but any future modification of those functions may have to be ported to these reimplementations.

QGarchery (approved):

src/MetaMorpho.sol
```diff
@@ -616,9 +616,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyCap = config[id].cap;
             if (supplyCap == 0) continue;
 
-            uint256 supplyAssets = MORPHO.expectedSupplyAssets(_marketParams(id), address(this));
```
MathisGD, line 619: shouldn't we rename this function in the morpho lib? I'm ready to bet that a lot of integrators will have the same issue

src/MetaMorpho.sol
```diff
@@ -772,7 +773,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             if (supplyCap == 0) continue;
 
             MarketParams memory marketParams = _marketParams(id);
-            (uint256 supplyAssets,,) = _accruedSupplyBalance(marketParams, id);
```
MathisGD, line 775: should this function be renamed?

MathisGD: why here we don't use the morpho balance lib?

Jean-Grimal: Which one ?

MerlinEgalite: because we round up in this contract while morpho balance lib rounds down

src/MetaMorpho.sol
QGarchery: What issue are you referring to ? Isn't this an issue to submit on the `morpho-blue` repo instead ?

Rubilmax: I see it really tied to the fact we need to under-estimate the maximum, so over-estimate every subtracted part to the maximum. Why would we create a specific function to over-estimate the supply assets of some account on Blue?

It seems very edge-casy

MerlinEgalite (approved):

src/MetaMorpho.sol
MathisGD: I meant why we were not using morpho lib

MathisGD (approved):

src/MetaMorpho.sol
```diff
@@ -616,9 +616,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 supplyCap = config[id].cap;
             if (supplyCap == 0) continue;
 
-            uint256 supplyAssets = MORPHO.expectedSupplyAssets(_marketParams(id), address(this));
+            uint256 supplyShares = MORPHO.supplyShares(id, address(this));
+            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) = MORPHO.expectedMarketBalances(_marketParams(id));
 
-            totalSuppliable += supplyCap.zeroFloorSub(supplyAssets);
+            totalSuppliable += supplyCap.zeroFloorSub(supplyShares.toAssetsUp(totalSupplyAssets, totalSupplyShares));
```
StErMi, line 622: Like you have done for `_supplyMorpho` maybe would be good to also add a comment here. I know that this is more a "utility" function and not a "core" one like `_supplyMorpho` but it will be probably used very much by integrators. 

Consider also adding it as a natspec `@dev` comment.

src/MetaMorpho.sol
StErMi: consider adding this comment also as a `@dev` natspec comment

StErMi:
I was worried about `if (assets != 0) revert ErrorsLib.AllCapsReached();` but both `_maxDeposit` and `_supplyMorpho` adopt the same behavior, so it should be fine as long as users use `maxDeposit` if they want to be sure to not revert because they try to supply above the cap.

PR #348, opened 2023-12-13
Delete pending cap when removing a market

Jean-Grimal:
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/ce072b25-742b-4a94-b430-4a24c445815b


test/forge/MarketTest.sol
```diff
@@ -179,6 +181,8 @@ contract MarketTest is IntegrationTest {
         assertEq(Id.unwrap(vault.withdrawQueue(1)), Id.unwrap(expectedWithdrawQueue[1]));
         assertEq(Id.unwrap(vault.withdrawQueue(2)), Id.unwrap(expectedWithdrawQueue[2]));
         assertFalse(vault.config(allMarkets[2].id()).enabled);
+        assertEq(vault.pendingCap(allMarkets[2].id()).value, 0, "Pending cap value");
```
Rubilmax, line 184: you can store `vault.pendingCap(allMarkets[2].id())` in memory ahah

Rubilmax:
```suggestion
        assertEq(vault.pendingCap(allMarkets[2].id()).value, 0, "pendingCap.value");
```

Rubilmax (approved):

MerlinEgalite (approved):

test/forge/MarketTest.sol
```diff
@@ -179,6 +181,8 @@ contract MarketTest is IntegrationTest {
         assertEq(Id.unwrap(vault.withdrawQueue(1)), Id.unwrap(expectedWithdrawQueue[1]));
         assertEq(Id.unwrap(vault.withdrawQueue(2)), Id.unwrap(expectedWithdrawQueue[2]));
         assertFalse(vault.config(allMarkets[2].id()).enabled);
+        assertEq(vault.pendingCap(allMarkets[2].id()).value, 0, "pendingCap.value");
+        assertEq(vault.pendingCap(allMarkets[2].id()).validAt, 0, "Pending cap validAt");
```
Rubilmax, line 185:
```suggestion
        assertEq(vault.pendingCap(allMarkets[2].id()).validAt, 0, "pendingCap.validAt");
```

Rubilmax (approved):

MerlinEgalite (approved):

MerlinEgalite (dismissed):

Rubilmax (approved):

MathisGD:
to me we should more prevent any cap setting when the market is being removed

MerlinEgalite:
> to me we should more prevent any cap setting when the market is being removed

indeed it sounds good

Rubilmax:
> to me we should more prevent any cap setting when the market is being removed

There's one thing for sure is that this PR does not 100% fix the issue raised: even if the pending cap is deleted, it can be submitted again afterwards
Am I right?

In which case I believe reverting when `removableAt != 0` in `_setCap` is safer

Rubilmax (changes requested):
See above

MerlinEgalite:
> > to me we should more prevent any cap setting when the market is being removed
> 
> There's one thing for sure is that this PR does not 100% fix the issue raised: even if the pending cap is deleted, it can be submitted again afterwards Am I right?
> 
> In which case I believe reverting when `removableAt != 0` in `_setCap` is better

which would be fixed by #353 right?

Jean-Grimal:
> > > to me we should more prevent any cap setting when the market is being removed
> > 
> > 
> > There's one thing for sure is that this PR does not 100% fix the issue raised: even if the pending cap is deleted, it can be submitted again afterwards Am I right?
> > In which case I believe reverting when `removableAt != 0` in `_setCap` is better
> 
> which would be fixed by #353 right?

I don't think so,

@MathisGD @Rubilmax should I make the change in this PR ?

src/MetaMorpho.sol
```diff
@@ -355,6 +355,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                 }
 
                 delete config[id];
+                delete pendingCap[id];
```
QGarchery, line 358: [optional] I realize now that these changes (both `config[id]` and `pendingCap[id]`) may be difficult to follow onchain with events. This requires to do the difference of sets to know which markets were removed. Maybe emitting an event here would be nice for offchain purposes

QGarchery (approved):

src/MetaMorpho.sol
MerlinEgalite: cc @julien-devatom

QGarchery (changes requested):
> There's one thing for sure is that this PR does not 100% fix the issue raised: even if the pending cap is deleted, it can be submitted again afterwards
Am I right?

I thought it would need happen after the removal of the market (or otherwise it gets deleted), which sounds like a wanted behavior.

But actually it could happen that the pending cap is accepted first in 2 cases:
- the removal of the market is done slowly after its timelock, for example a couple of blocks slip by and the pending cap timelock is reached
- the timelock is reduced, and a new pending cap is submitted before the end of the removal of the market timelock

So in the end I agree that we should enforce stricter rules on the pending cap

Rubilmax (changes requested):
This PR is complex to review aside from #353 

I would merge #353 first and then make this PR conform to this new standard behavior (prefer reverting over having side effects)

src/MetaMorpho.sol
```diff
@@ -355,6 +356,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
```
MathisGD, line 359: shouldn't we revert instead of removing the cap ?

Rubilmax: This would be consistent with #353 indeed. I would wait for it to be finalized and merged though

src/MetaMorpho.sol
```diff
@@ -354,6 +354,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                     }
                 }
 
+                if (pendingCap[id].validAt != 0) revert ErrorsLib.PendingCap(id);
```
MerlinEgalite, line 357: let's revert earlier no?

MerlinEgalite (changes requested):

src/MetaMorpho.sol
MathisGD: yes, next to this

```solidity
if (config[id].cap != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroCap(id);
```

src/libraries/ErrorsLib.sol
```diff
@@ -43,7 +43,7 @@ library ErrorsLib {
     error AlreadyPending();
 
     /// @notice Thrown when submitting the removal of a market when there is a cap already pending on that market.
-    error PendingCap();
+    error PendingCap(Id id);
```
MathisGD, line 46: can we have both (one with and one without) ?

Jean-Grimal: Yes we could, but I guess that having only one works fine no ?

src/MetaMorpho.sol
```diff
@@ -737,6 +738,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         MarketConfig storage marketConfig = config[id];
 
         if (supplyCap > 0) {
+            if (config[id].removableAt != 0) revert ErrorsLib.PendingRemoval();
```
MathisGD, line 741: we don't need this as it is checked in submit cap

Jean-Grimal: Yes but if I remember well we wanted to also have it here as a security.

MerlinEgalite (dismissed):

src/MetaMorpho.sol
MathisGD: hmm feels weird

src/MetaMorpho.sol
```diff
@@ -294,7 +294,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     function submitMarketRemoval(MarketParams memory marketParams) external onlyCuratorRole {
         Id id = marketParams.id();
         if (config[id].removableAt != 0) revert ErrorsLib.AlreadyPending();
-        if (pendingCap[id].validAt != 0) revert ErrorsLib.PendingCap();
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.PendingCap(id);
```
StErMi, line 297: For a better readability, it would be better to have all the `config` checks done one after another and the `pendingCap` one at the very end.

Also, it should avoid an `SLOAD` if something reverts because of the checks done within the `config`

StErMi:
https://github.com/morpho-org/metamorpho/pull/348/files#diff-654c1b2b88c5379e79b25c1c0df42db121f692b705307dc057d5808f86b5ea8dR358

Should you also delete any pending cap proposal when the market is removed from the withdrawal queue?

src/MetaMorpho.sol
```diff
@@ -345,6 +345,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                 Id id = withdrawQueue[i];
 
                 if (config[id].cap != 0) revert ErrorsLib.InvalidMarketRemovalNonZeroCap(id);
+                if (pendingCap[id].validAt != 0) revert ErrorsLib.PendingCap(id);
```
StErMi, line 348: Why the revert for this condition?

MerlinEgalite: There are 2 ways to remove a markets: 
1. Either there's no supply on it, the cap has been set to 0 previously and the market can be safely removed by triggering `updateWithdrawQueue`.
2. Either there's some supply on it. In this case, the cap must be set to 0, `submitMarketRemoval` must be triggered. After the timelock has passed `updateWithdrawQueue` can be triggered.

MerlinEgalite: So this check make sure that no pending cap has been set to a positive value while the cap is now at 0. It enforces to revoke the pending value to make sure the state is clean and no market will be readded right after being removed.

src/MetaMorpho.sol
MerlinEgalite: Indeed we could, even if it's not a big deal IMO

StErMi: You know... it's a nitpick 😎

MerlinEgalite:
> Should you also delete any pending cap proposal when the market is [removed from the withdrawal queue](https://github.com/morpho-org/metamorpho/pull/348/files#diff-654c1b2b88c5379e79b25c1c0df42db121f692b705307dc057d5808f86b5ea8dR358)?

I don't think so since `pendingCap[id].validAt` is checked before. If this timestamp is null then the pending is null as well.

QGarchery (approved):
Sounds good, it fixes [this comment](https://github.com/morpho-org/metamorpho/pull/348#pullrequestreview-1794597944) because now the pending cap prevents removal (which means there are no problem of ordering market removal and updating caps)

src/MetaMorpho.sol
MerlinEgalite: @Jean-Grimal can you update the PR please?

MathisGD: pushed

MathisGD (dismissed):

MathisGD (approved):

StErMi (approved):

MerlinEgalite (approved):

PR #349, opened 2023-12-13
Update `post-cantina`

Rubilmax:


MerlinEgalite (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -30,57 +30,148 @@ interface IOwnable {
...
+
+    /// @notice The skim recipient.
     function skimRecipient() external view returns (address);
+
+    /// @notice The current timelock.
     function timelock() external view returns (uint256);
+
+    /// @dev Stores the order of markets on which liquidity is supplied upon deposit.
+    /// @dev Can contain any market. A market is skipped as soon as its supply cap is reached.
     function supplyQueue(uint256) external view returns (Id);
```
StErMi, line 59: Does not have `@notice`, `@param` and `@return`

I would have anyway suggested documenting it properly, but I think that in this case is even more important given that the name of the function seems more "active" (by executing you supply to the queue) while instead it's a getter.

src/interfaces/IMetaMorpho.sol
```diff
@@ -30,57 +30,148 @@ interface IOwnable {
...
+
+    /// @notice Returns the length of the supply queue.
     function supplyQueueLength() external view returns (uint256);
+
+    /// @dev Stores the order of markets from which liquidity is withdrawn upon withdrawal.
+    /// @dev Always contain all non-zero cap markets as well as all markets on which the vault supplies liquidity,
+    /// without duplicate.
     function withdrawQueue(uint256) external view returns (Id);
```
StErMi, line 67: Does not have `@notice`, `@param` and `@return`

I would have anyway suggested documenting it properly, but I think that in this case is even more important given that the name of the function seems more "active" (by executing you supply to the queue) while instead it's a getter.

src/interfaces/IMetaMorpho.sol
StErMi: The `supplyQueue`, as far as I remember, can contain duplicates (I don't think you have changed such behavior).

In the `withdrawQueue` you specify that it does not contain duplicates. You should also specify with a `@dev` comment that `supplyQueue` **can** contain duplicates.

src/interfaces/IMetaMorpho.sol
```diff
@@ -30,57 +30,148 @@ interface IOwnable {
...
+
+    /// @notice Accepts the pending cap of the market defined by `id`.
     function acceptCap(Id id) external;
+
+    /// @notice Revokes the pending cap of the market defined by `id`.
     function revokePendingCap(Id id) external;
 
+    /// @notice Submits a forced market removal from the vault, potentially losing all funds supplied to the market.
+    /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
     function submitMarketRemoval(Id id) external;
```
StErMi, line 100: It could be worth specifying that this function does set the supply cap to `0` immediately, but does not remove such market from the withdraw queue. You need to later on call `updateWithdrawQueue` to remove it from the queue.

src/interfaces/IMetaMorpho.sol
```diff
@@ -30,57 +30,148 @@ interface IOwnable {
...
     function submitGuardian(address newGuardian) external;
+
+    /// @notice Accepts the pending guardian.
     function acceptGuardian() external;
+
+    /// @notice Revokes the pending guardian.
     function revokePendingGuardian() external;
 
+    /// @notice Skims the vault `token` balance to `skimRecipient`.
     function skim(address) external;
```
StErMi, line 119: It's missing the parameter name in the signature

src/interfaces/IMetaMorpho.sol
```diff
@@ -30,57 +30,148 @@ interface IOwnable {
 /// @dev This interface is used for factorizing IMetaMorphoStaticTyping and IMetaMorpho.
 /// @dev Consider using the IMetaMorpho interface instead of this one.
 interface IMetaMorphoBase {
```
StErMi, line 32: @MerlinEgalite please verity that **each** function is properly and fully natspec documented with at least `@notice` and all the needed `@param` and `@return` tags. Many of the functions are missing these tags.

src/interfaces/IMetaMorphoFactory.sol
```diff
@@ -0,0 +1,32 @@
...
+    function isMetaMorpho(address target) external view returns (bool);
+
+    /// @notice Creates a new MetaMorpho vault.
+    /// @param initialOwner The owner of the vault.
+    /// @param initialTimelock The initial timelock of the vault.
+    /// @param asset The address of the underlying asset.
+    /// @param name The name of the vault.
+    /// @param symbol The symbol of the vault.
+    /// @param salt The salt to use for the MetaMorpho vault's CREATE2 address.
+    function createMetaMorpho(
```
StErMi, line 24: is missing the `@return` tag

PR #351, opened 2023-12-15
Remove automatic supply queue push

### LINKED ISSUE #352: `setSupplyQueue` and `setWithdrawQueue` are not emitted in `setCap`
https://github.com/morpho-org/metamorpho/blob/0b63c348445fea57532e3dd5d4962961d770d5a6/src/libraries/EventsLib.sol#L64-L68

MathisGD:
https://morpholabs.slack.com/archives/C02N7CSP91C/p1702569747842329

and fixes #352

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -743,6 +743,8 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
                 if (withdrawQueue.length > ConstantsLib.MAX_QUEUE_LENGTH) revert ErrorsLib.MaxQueueLengthExceeded();
 
                 marketConfig.enabled = true;
+
+                emit EventsLib.SetWithdrawQueue(msg.sender, withdrawQueue);
             }
 
             marketConfig.removableAt = 0;
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -738,15 +738,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
         if (supplyCap > 0) {
             if (!marketConfig.enabled) {
-                supplyQueue.push(id);
                 withdrawQueue.push(id);
 
-                if (
-                    supplyQueue.length > ConstantsLib.MAX_QUEUE_LENGTH
-                        || withdrawQueue.length > ConstantsLib.MAX_QUEUE_LENGTH
-                ) {
-                    revert ErrorsLib.MaxQueueLengthExceeded();
-                }
+                if (withdrawQueue.length > ConstantsLib.MAX_QUEUE_LENGTH) revert ErrorsLib.MaxQueueLengthExceeded();
 
                 marketConfig.enabled = true;
             }
--- a/test/forge/MetaMorphoInternalTest.sol
+++ b/test/forge/MetaMorphoInternalTest.sol
@@ -31,6 +31,7 @@ contract MetaMorphoInternalTest is InternalTest {
 
         Id id = allMarkets[0].id();
         _setCap(id, CAP);
+        supplyQueue = [id];
 
         loanToken.setBalance(SUPPLIER, suppliedAmount);
         vm.prank(SUPPLIER);
--- a/test/forge/helpers/IntegrationTest.sol
+++ b/test/forge/helpers/IntegrationTest.sol
@@ -100,7 +100,7 @@ contract IntegrationTest is BaseTest {
     function _setCap(MarketParams memory marketParams, uint256 newCap) internal {
         Id id = marketParams.id();
         uint256 cap = vault.config(id).cap;
-        bool wasEnabled = vault.config(id).enabled;
+        bool isEnabled = vault.config(id).enabled;
         if (newCap == cap) return;
 
         PendingUint192 memory pendingCap = vault.pendingCap(id);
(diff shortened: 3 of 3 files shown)


test/forge/helpers/IntegrationTest.sol
```diff
@@ -115,6 +116,18 @@ contract IntegrationTest is BaseTest {
...
+            if (!wasEnabled) {
+                Id[] memory newSupplyQueue = new Id[](vault.supplyQueueLength() + 1);
+                for (uint256 k; k < vault.supplyQueueLength(); k++) {
+                    newSupplyQueue[k] = vault.supplyQueue(k);
+                }
+                newSupplyQueue[vault.supplyQueueLength()] = id;
+                vm.prank(ALLOCATOR);
+                vault.setSupplyQueue(newSupplyQueue);
+            }
+        }
```
Jean-Grimal, line 130: Couldn't we have a dedicated function that adds a market to the supply queue ?

Jean-Grimal (approved):

test/forge/helpers/IntegrationTest.sol
MathisGD: if you want but it's only used here so Im' not sure that it's useful

src/MetaMorpho.sol
```diff
@@ -738,17 +738,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
-                    supplyQueue.length > ConstantsLib.MAX_QUEUE_LENGTH
-                        || withdrawQueue.length > ConstantsLib.MAX_QUEUE_LENGTH
-                ) {
-                    revert ErrorsLib.MaxQueueLengthExceeded();
-                }
+                if (withdrawQueue.length > ConstantsLib.MAX_QUEUE_LENGTH) revert ErrorsLib.MaxQueueLengthExceeded();
 
                 marketConfig.enabled = true;
+
+                emit EventsLib.SetWithdrawQueue(msg.sender, withdrawQueue);
```
Rubilmax, line 747: I am not sure it is a real issue
But if it is, we could also create a dedicated issue `AppendWithdrawQueue` to decrease the gas cost (because this line alone is responsible for 1 event + N cold SLOADs = 1k + N * 2.1k)

Rubilmax: @julien-devatom

test/forge/helpers/IntegrationTest.sol
```diff
@@ -100,6 +100,7 @@ contract IntegrationTest is BaseTest {
     function _setCap(MarketParams memory marketParams, uint256 newCap) internal {
         Id id = marketParams.id();
         uint256 cap = vault.config(id).cap;
+        bool wasEnabled = vault.config(id).enabled;
```
MerlinEgalite, line 103:
```suggestion
        bool isEnabled = vault.config(id).enabled;
```

test/forge/MetaMorphoInternalTest.sol
```diff
@@ -31,6 +31,7 @@ contract MetaMorphoInternalTest is InternalTest {
 
         Id id = allMarkets[0].id();
         _setCap(id, CAP);
+        supplyQueue = [id];
```
MerlinEgalite, line 34: why do we need this?

MathisGD: because here we are in metamorpho itself so to replicate the previous behavior we do this

test/forge/helpers/IntegrationTest.sol
Rubilmax: Indeed overkill for now but perhaps later

src/MetaMorpho.sol
Rubilmax: Seen w @julien-devatom , it seems fine to have another event `AppendWithdrawQueue` to limit gas cost if we want to do so

MerlinEgalite: whatever fits me tbh

Rubilmax (approved):

Rubilmax (changes requested):
this discussion needs to be resolved first: https://github.com/morpho-org/metamorpho/pull/351/files#r1427738018

Rubilmax (approved):

adhusson (approved):
This also fixes MM-3 from my report (but duplicates can still be set manually, and so `maxDeposit` can still be wrong).

Jean-Grimal (approved):

test/forge/helpers/IntegrationTest.sol
```diff
@@ -115,6 +116,18 @@ contract IntegrationTest is BaseTest {
         vault.acceptCap(id);
 
         assertEq(vault.config(id).cap, newCap, "_setCap");
+
+        if (newCap > 0) {
+            if (!isEnabled) {
```
QGarchery, line 121:
```suggestion
        if (newCap > 0 && !isEnabled) {
```

QGarchery (approved):

MerlinEgalite (approved):

StErMi:
The PR seems OK. Consider documenting the de-sync between the `withdrawQueue` and `supplyQueue` (before, when a market was added, both the supply and withdraw queue were configured and pushed).

You could have the whole `withdrawQueue` fully configured, but no market from which you can supply to.

QGarchery:
> You could have the whole `withdrawQueue` fully configured, but no market from which you can supply to.

Yes this is a wanted behavior: some vault curators will prefer reallocating every time instead of using the supply queue.

> Consider documenting the de-sync between the withdrawQueue and supplyQueue (before, when a market was added, both the supply and withdraw queue were configured and pushed).

I think it's ok considering that there is already other functions that could desync the two queues (which seems natural since the code has 2 queues, so it would make sense if they are not always the same)

PR #353, opened 2023-12-15
Always revert when there is a value already pending

### LINKED ISSUE #221: [protocol review] Possible accidental removal of pending transactions
The deletion of pending values happen as soon as the current value is updated, whether or not the pending value has actually been used. This means that some awaited change can be accidentally removed by the owner of the vault.
Consider the following scenario:
1. a vault owner wants to plan high fees in the future, and as a counterparty low fees right now
2. the vault owner `submit` high fees, this will be subject to a timelock as expected
3. the vault owner `submit` lower fees, this is effective immediately
4. the vault owner will not be able to `acceptFee` in the future because the `pendingFee` has been deleted by the `_setFee` subroutine

MathisGD:
https://morpholabs.slack.com/archives/C02N7CSP91C/p1702634824993509

Fixes #221


src/MetaMorpho.sol
```diff
@@ -208,14 +208,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IMetaMorphoBase
     function submitTimelock(uint256 newTimelock) external onlyOwner {
         if (newTimelock == timelock) revert ErrorsLib.AlreadySet();
+        if (pendingTimelock.validAt != 0) revert ErrorsLib.AlreadyPending();
```
Rubilmax, line 211: This design is less prone to error for sure, but is worse in terms of UX: you have to batch a `revokePendingTimelock` with a `submitTimelock`

I also think it's fine, but I believe it is the reason we didn't do this earlier

src/MetaMorpho.sol
```diff
@@ -273,18 +268,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IMetaMorphoBase
     function submitCap(MarketParams memory marketParams, uint256 newSupplyCap) external onlyCuratorRole {
         Id id = marketParams.id();
+        uint256 supplyCap = config[id].cap;
         if (marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
         if (MORPHO.lastUpdate(id) == 0) revert ErrorsLib.MarketNotCreated();
-
-        uint256 supplyCap = config[id].cap;
```
Rubilmax, line 279: Why moving up the SLOAD?
`supplyCap` is not used until L280

Rubilmax: By the same rule, id should be assigned 1 line later

Rubilmax (approved):
This design relies on the fact the pending value is deleted every time a value is set

src/MetaMorpho.sol
Rubilmax: I would block this PR bnecause of this

MerlinEgalite: I'd be in favor of reverting back this change as well

src/MetaMorpho.sol
```diff
@@ -273,18 +268,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.AlreadyPending();
         if (newSupplyCap == supplyCap) revert ErrorsLib.AlreadySet();
 
         if (newSupplyCap < supplyCap) {
             _setCap(id, newSupplyCap.toUint184());
         } else {
-            // newSupplyCap > supplyCap >= 0 so there's no need to check `pendingCap[id].validAt != 0`.
-            if (newSupplyCap == pendingCap[id].value) revert ErrorsLib.AlreadyPending();
```
MerlinEgalite, line 286: we should keep reverting at this place IMO for the reason mentioned above

src/MetaMorpho.sol
MerlinEgalite: I think it can be more problematic for the cap when the curator or owner want to act fast to disable a market and an updated is pending...

MerlinEgalite (changes requested):

Jean-Grimal (changes requested):
The natspec should be updated

src/MetaMorpho.sol
Rubilmax: But they can bundle a revoke + submit so they can essentially act "fast" (within a block)

MerlinEgalite: Yes ofc but this must be clear for them, else you can easily lost minutes because of that, so you need to have something in the fronted and in the doc at least to cover this specific case

src/MetaMorpho.sol
Rubilmax: I don't get this: this check is weaker than the one introduced, why would we keep it?
Meaning that if the check introduced passes, this one should pass too

Rubilmax: I don't get it: if the check introduced passes (no pending value), then this one should not pass (the new value cannot equal the pending value)
Why would we keep this check?

src/MetaMorpho.sol
MathisGD: we don't usually for gas cost of reverts right?

Rubilmax: Yes but also why assign variables too early (wrt when they are actually used)?

src/MetaMorpho.sol
MerlinEgalite: This enforces [this](https://github.com/morpho-org/metamorpho/pull/353#discussion_r1428051554). Which I think is very annoying... In case we still want to push that change, we must provide better explanations on the process of setting a cap to 0 in case of emergency in case a pending value has been submitted recently.

Rubilmax (changes requested):
Pointing out that being consistent with #348 requires to revert in `submitMarketRemoval` and `submitCap`

src/MetaMorpho.sol
MathisGD: I find it ugly and useless but ok if you want I'm changing

src/MetaMorpho.sol
Rubilmax: Alright so I wouldn't argue that we should keep this check but instead move the check introduced in place of this check, so we keep the feature that bypasses the timestamp

It indeed feels weird to have the feature to bypass the timestamp but only when there's not already a pending value
But it's not that weird either: it's like having a pending value that goes in one direction but still being able to instantly push a value in the other direction

src/MetaMorpho.sol
```diff
@@ -286,10 +286,10 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IMetaMorphoBase
     function submitMarketRemoval(Id id) external onlyCuratorRole {
         if (config[id].removableAt != 0) revert ErrorsLib.AlreadyPending();
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.PendingCap();
+        if (config[id].cap != 0) revert ErrorsLib.NonZeroCap();
```
MerlinEgalite, line 290: Now the process to remove a market when there's a pending cap is:
1. `revokePendingCap`
2. `submitCap(0)`
3. `submitMarketRemoval()`

This must clearly explained in the README IMO

src/MetaMorpho.sol
adhusson: I'm also leaning towards having a special case and allowing setting a cap to 0 even with a pending cap. In practice there is a small difference between submitting a call and submitting a bundle (and between verifying each).

Jean-Grimal (approved):
A test on submitting a guardian while there is already a pending guardian should be added

src/MetaMorpho.sol
QGarchery: I agree that it should be documented and that setting a cap to 0 should be made easy, but I would not go as far as introducing a special case for it. Adding a different operational method for handling pending values is also a mental overhead.

As @Rubilmax said, I don't find it so weird to have to revoke a pending value if the admin wants to actually go in the other direction of the pending value

src/MetaMorpho.sol
```diff
@@ -275,16 +270,13 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         Id id = marketParams.id();
         if (marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
         if (MORPHO.lastUpdate(id) == 0) revert ErrorsLib.MarketNotCreated();
-
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.AlreadyPending();
```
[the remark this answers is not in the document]
MathisGD: yes very good idea

src/MetaMorpho.sol
QGarchery, line 1: I just realized that with this change there is no multicall that guarantees that some operations that need to be done fast (set 0 cap, submit forced market removal) go through. For example, there can be a pending cap, so a bundle with a revoke of the pending cap may seem necessary. But the guardian can also frontrun (intentionally or not) and revoke the pending cap, which means that the bundle would revert.

It seems annoying because we don't want to have to develop another contract to handle those cases, or force curators of different MM vaults to develop it themselves. 

A fix for that would be to allow revoke transactions to go through, even if they cause no change

QGarchery (changes requested):

src/MetaMorpho.sol
adhusson: I like the suggestion. And if some functions are modified and don't revert even when they cause no change they could be renamed to `ensureX` (like `ensurePendingTimelockRevoked`) so the difference in behaviour is clear.

In general I don't really understand what concrete issues are prevented by "revert `setVal(x)` if `val` is already `x`" checks, or by this revert-if-already-pending check.

Jean-Grimal (dismissed):

src/MetaMorpho.sol
MerlinEgalite: Note that if it reverts once you can just update the bundler to make it goes through so it's not that problematic. I mean it's not like the guardian can totally prevent the action. In the end I'm ambivalent.

MerlinEgalite: Do you request the change until we apply you suggestion @QGarchery ?

MathisGD: Classically in multicall contracts is there a way to say that you don't want to revert if one of the calls reverts?

Otherwise I'm ok with your proposal Quentin, with the right naming change

Jean-Grimal (dismissed):

src/MetaMorpho.sol
adhusson: I think the issue is that speed might be important so you want to have a single tx ready to go that ensures some condition holds.

MerlinEgalite: > Classically in multicall contracts is there a way to say that you don't want to revert if one of the calls reverts?

In multicall used no

adhusson:
After going over the abilities of each roles, the following are candidate for "must be doable in one prepared bundle, no frontrun risk". I think whoever submits the bundle should not even have to select among multiple bundles depending on the current state. For instance there should be a single "set cap to 0" bundle that always works.

1. Lower supply cap (curator, owner) (noted by @QGarchery)

    This is important enough that the cap update code explicitly bypasses the timelock when the update lowers the cap.

    * As long as `revokePendingCap` reverts when there is no pending cap (it is the case right now), the check `validAt != 0` should only be performed when the new cap is larger. 

    * If `revokePendingMarketRemoval` reverts when there is no pending market removal (it is not the case right now), the check `removableAt != 0` should only be performed only when the new cap is larger.

2. Revoke a bad pending value (owner, guardian, + curator for `revokePendingCap`)

    Currently fine, will only revert of the value is already revoked or (for `revokePendingMarketRemoval`) will not revert.

src/MetaMorpho.sol
MerlinEgalite: what is the appropriate naming @MathisGD for you? `revokePendingMarketRemoval` has already the behavior suggested by @QGarchery without naming difference

MerlinEgalite: https://github.com/morpho-org/metamorpho/pull/366

adhusson (dismissed):

adhusson (approved):

adhusson (approved):

MerlinEgalite (approved):
I still think the naming change was not necessary

src/MetaMorpho.sol
```diff
@@ -424,27 +417,21 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /* REVOKE FUNCTIONS */
 
     /// @inheritdoc IMetaMorphoBase
-    function revokePendingTimelock() external onlyGuardianRole {
-        if (pendingTimelock.validAt == 0) revert ErrorsLib.NoPendingValue();
-
+    function revokePendingTimelockNoRevert() external onlyGuardianRole {
```
Rubilmax, line 420: this naming is very weird from an integrator POV: one may expect this function to never revert but in fact it can revert for any account that doesn't have the guardian role
I understand the motivation but I don't think naming it this way is the way to go, because i don't think it is even useful

Just mention it in the spec of the function and that's enough

Also:
1. `revokePendingMarketRemoval` is not consistent with this naming
2. why not name `acceptCapNoRevert` etc?

Rubilmax (changes requested):

src/MetaMorpho.sol
MathisGD: Agree that this naming isn't the best. I think though that we shouldn't stay with the previous name, because the behavior isn't consistent with the rest of the contracts. What do you think about  ˋremovePendingXIfAny`?

Rubilmax: I don't like `revokePendingXIfAny` neither, though I prefer it over the current solution

It is verbose to name a function like this and inconsistent with the rest (see my question: why not name `acceptCapNoRevert`?) while the benefit is to point out missing requirements (removed on purpose)

If I'm not mistaken: we're removing the requirement because we don't want to revert in a multicall bundle. It's like when we've removed the role requirement of `acceptX` functions because we want anybody to be able to accept a change, but yet we didn't renamed functions `acceptXPublic` or `acceptXNoRole`

Rubilmax: In other words, the requirement is removed from this function on purpose, because this function's specs have changed ; but we don't want to name functions so they have the most descriptive and specific name per their spec

adhusson: After thinking a bit I changed my mind you're right. I think there was a strong argument for it when one of the revoke functions did not work like the others (it did not have the same requirement) but now it's ok to leave as is.

MerlinEgalite:
@MathisGD can you revert back the naming and fix the conflicts please?

Jean-Grimal (approved):

MerlinEgalite (approved):

src/MetaMorpho.sol
```diff
@@ -214,14 +214,12 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IMetaMorphoBase
     function submitTimelock(uint256 newTimelock) external onlyOwner {
         if (newTimelock == timelock) revert ErrorsLib.AlreadySet();
+        if (pendingTimelock.validAt != 0) revert ErrorsLib.AlreadyPending();
```
StErMi, line 217: Before, you could have a pending value (decrease of timelock) that could have been deleted by another `submitTimelock` that set the new timelock to a new one.

Now you won't be able to do that, and you are forced to wait for the timelock to expire or revoke it and call it again.

Is this the intended new behavior?

src/MetaMorpho.sol
```diff
@@ -281,16 +276,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         Id id = marketParams.id();
         if (marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
         if (MORPHO.lastUpdate(id) == 0) revert ErrorsLib.MarketNotCreated();
-
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.AlreadyPending();
+        if (config[id].removableAt != 0) revert ErrorsLib.PendingRemoval();
```
StErMi, line 280: I think that this is a welcome check because the removal of a market is a very strong operation that also has a timelock that is different compared to the cap timelock (and previously they were non correlated within the logic). Now if a market is marked as "to-be-removed" such operation has the priority and will lock any changes to the cap.

Now as soon as a market is marked as to-be-removed, there's no way to re-enable the cap (that is already set to zero) unless you explicitly call `revokePendingMarketRemoval`

src/MetaMorpho.sol
```diff
@@ -281,16 +276,14 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         Id id = marketParams.id();
         if (marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
         if (MORPHO.lastUpdate(id) == 0) revert ErrorsLib.MarketNotCreated();
-
+        if (pendingCap[id].validAt != 0) revert ErrorsLib.AlreadyPending();
```
StErMi, line 279: Like for the `submitTimelock`, this new check brings some side effects. Unless you revoke the previous pending proposal, or you wait for the timelock + accept it, you cannot

- create a new proposal
- insta-change the supply cap (when the new cap is lower)

In general, you are giving much more priority and power to the pending proposals. Is this the intended behavior?

StErMi:
Would you mind confirming the logic of the behavior? Because otherwise, I'm trying to reverse engineering it and could arrive at the wrong conclusion.

The new logic allows the submission of a change to the

- timelock
- guardian
- supply
- market removal

Only if there's no direct or indirect pending proposal currently in action.

To be more clear
- direct: you can't submit a cap proposal if there's already one
- indirect: you can't submit a cap proposal if there's a pending market removal proposal

Moreover, there are some "custom" requirements (that still need to be documented), like for example

- you can't require a market removal if the cap is not already set to zero. In general, you don't want to execute implicit logic, but the state must be already "ready" to perform the action once the timelock has ended.

Would you mind confirming or detailing more/correct my recap?

StErMi:
Giving the immutable nature of MM (you can't upgrade and change the logic) it could make sense to create utility contracts to manage it. 

Like for example an external contract (with the owner or curator role) that perform all the actions needed to be able to execute `MM.submitMarketRemoval`

- if the cap is not 0 and there's a pending proposal, revoke it and set it to zero
- call the `submitMarketRemoval`

src/MetaMorpho.sol
MerlinEgalite: yes we enforce to revoke every pending values before submitting a new one.

src/MetaMorpho.sol
MerlinEgalite: yes it is. What we'll do is that in the frontend we can leverage the multicall functionality to bundle a revoke + submit in a single tx to perform the required actions.

MerlinEgalite:
I confirm your first comment. We wanted `submit*`, `revoke*` and `accept*` functions to be single purpose functions with less side effects than previously.


For the second comment, the vault inherits from `Multicall` which allows to extract that logic offchain, so what you proposed will be done in Morpho Blue's frontend. Nonetheless, some vault curators are likely to write their own logic but we won't write it ourselves.

PR #354, opened 2023-12-18
docs(metamorpho): comment forced market removal

Rubilmax:


MerlinEgalite (dismissed):

adhusson (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -95,8 +95,12 @@ interface IMetaMorphoBase {
     /// @notice Revokes the pending cap of the market defined by `id`.
     function revokePendingCap(Id id) external;
 
-    /// @notice Submits a forced market removal from the vault, potentially losing all funds supplied to the market.
+    /// @notice Submits a forced market removal from the vault, losing all funds supplied to the market.
```
Jean-Grimal, line 98: I think we should keep "potentially" as there might be a chance that force removed market can be added again, and in that case the funds aren't lost.

It doesn't work if the market is doomed forever, but if working markets are removed (because of a malicious curator), the funds are not definitely lost.

Jean-Grimal (changes requested):

src/interfaces/IMetaMorpho.sol
MerlinEgalite, line 98:
```suggestion
    /// @notice Submits a forced market removal from the vault, losing all funds supplied to the market unless the market is re-enabled later on.
```

Rubilmax: But I don't like this because it's pointing out funds can be recovered later on except it funds will be distributed pro-rata to the shares when recovered, not pro-rata to the shares when lost...

Rubilmax: https://github.com/morpho-org/metamorpho/pull/354#discussion_r1433753164

MerlinEgalite (dismissed):

Jean-Grimal (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -95,8 +95,14 @@ interface IMetaMorphoBase {
...
+    /// @notice Funds can always be recovered as a donation by withdrawing from this market (using `reallocate`), but
```
QGarchery, line 99: Beware that this line conflicts with #358

src/interfaces/IMetaMorpho.sol
```diff
@@ -95,8 +95,14 @@ interface IMetaMorphoBase {
...
+    /// funds will be distributed pro-rata to the shares at the time of withdrawal, not at the time of market removal.
+    /// @notice This forced removal is expected to be used as an emergency process in case a market constantly reverts.
+    /// To softly remove a sane market, the curator role is expected to bundle a reallocation that empties the market
+    /// first (using `reallocate`), followed by the removal of the market (using `updateWithdrawQueue`).
     /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
+    /// @dev Warning: Removing a market with non-zero supply will instantly impact the vault's price per share.
```
MathisGD, line 105: nb: donating to the vault too

MathisGD (approved):

MathisGD (dismissed):

src/interfaces/IMetaMorpho.sol
QGarchery: Since it seems like we plan to merge #358, we can base this PR on top of it

src/interfaces/IMetaMorpho.sol
```diff
@@ -96,6 +96,13 @@ interface IMetaMorphoBase {
     function revokePendingCap(Id id) external;
 
     /// @notice Submits a forced market removal from the vault, eventually losing all funds supplied to the market.
+    /// @notice Funds can always be recovered as a donation by withdrawing from this market (using `reallocate`), but
+    /// funds will be distributed pro-rata to the shares at the time of withdrawal, not at the time of market removal.
+    /// @notice This forced removal is expected to be used as an emergency process in case a market constantly reverts.
+    /// To softly remove a sane market, the curator role is expected to bundle a reallocation that empties the market
+    /// first (using `reallocate`), followed by the removal of the market (using `updateWithdrawQueue`).
+    /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
```
QGarchery, line 104: I would remove this line now that we check that `removableAt` is 0 at the beginning of this function

src/interfaces/IMetaMorpho.sol
QGarchery: And bad debt realization on an enabled market too, but it's ok to not mention it here because they are not directly related to this `submitMarketRemoval` function

QGarchery (changes requested):

Jean-Grimal (dismissed):

MathisGD (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):

PR #355, opened 2023-12-18
Revert when submitting cap and a removal has been submitted

Jean-Grimal:


src/MetaMorpho.sol
```diff
@@ -275,6 +275,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         Id id = marketParams.id();
         if (marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
         if (MORPHO.lastUpdate(id) == 0) revert ErrorsLib.MarketNotCreated();
+        if (config[id].removableAt != 0) revert ErrorsLib.RemovalSubmitted();
```
MathisGD, line 278:
```suggestion
        if (config[id].removableAt != 0) revert ErrorsLib.PendingRemoval();
```

src/libraries/ErrorsLib.sol
```diff
@@ -30,6 +30,9 @@ library ErrorsLib {
     /// asset.
     error InconsistentAsset(Id id);
 
+    /// @notice Thrown when submitting a cap for a market `id` which removal has been submitted.
+    error RemovalSubmitted();
```
Rubilmax, line 34: It is indeed clearer to have a dedicated error for this case

src/libraries/ErrorsLib.sol
```diff
@@ -30,6 +30,9 @@ library ErrorsLib {
     /// asset.
     error InconsistentAsset(Id id);
 
+    /// @notice Thrown when submitting a cap for a market `id` which removal has been submitted.
```
MathisGD, line 33:
```suggestion
    /// @notice Thrown when submitting a cap for a market `id` with a pending removal.
```

src/MetaMorpho.sol
Rubilmax, line 278: I believe being consistent with the current code would require this check to be moved in the `else` branch below

Like the curator role can always instantly set a new cap, even if the market is submitted for removal

The cap being set at 0 upon removal submission makes this unnecessary, because the curator role can only increase the cap from here

So the only benefit would be 1 less SLOAD involved when curator role wants to instantly lower a cap on a market

adhusson (approved):

src/MetaMorpho.sol
Jean-Grimal: Yes I agree

MerlinEgalite (approved):

MathisGD (approved):

PR #347, opened 2023-12-12
Fix _withdraw() natspec

Jean-Grimal:
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/637922bd-c38b-4ea0-bb51-a1420a47d088

PR #345, opened 2023-12-12
Update doc and natspec

Jean-Grimal:
Fixes :
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/a29d3775-3516-4013-b78a-69d00f0dc43a
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/3442e8f8-d0a2-46bd-83a8-67208038e5f5
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/1d74afc9-25a1-4c51-a004-fbaa4ee8f3e3
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/47b33a14-c10e-41cf-921e-19236493856a
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/0ce3ec22-ab52-4e6a-afed-ab98a7bc1fde
- https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/6a478755-6285-42fe-b0e9-04eb43ba6f2f

PR #344, opened 2023-12-12
docs(fee): add acknowledegement

Rubilmax:
- Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/dc29a0e6-64b1-4f9b-a73e-9ce28b7f34a4

PR #343, opened 2023-12-12
Add comment on reallocate

Jean-Grimal:
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/63dbc883-ecf0-4a6a-a844-0cfb015127c9

PR #357, opened 2023-12-19
Post cantina

MerlinEgalite:
Contains:

- https://github.com/morpho-org/metamorpho/pull/349 : Fix : moves natspec to interfaces.
- https://github.com/morpho-org/metamorpho/pull/350
- https://github.com/morpho-org/metamorpho/pull/341 : `Owner()` function was not marked as view.
- https://github.com/morpho-org/metamorpho/pull/344 : Fix : Add acknowledgment on fees that might be rounded to zero.
- https://github.com/morpho-org/metamorpho/pull/351 : Fix : remove automatic push in supply queue when enabling a new market.
- https://github.com/morpho-org/metamorpho/pull/345
- https://github.com/morpho-org/metamorpho/pull/343 : Fix : Add of a recommandation in the `reallocate` function natspec to help not revert.
- https://github.com/morpho-org/metamorpho/pull/360
- https://github.com/morpho-org/metamorpho/pull/339 : reduce gas by passing `marketParams` instead of `id`
- https://github.com/morpho-org/metamorpho/pull/347 : Wrong natspec for `_withdraw()`
- https://github.com/morpho-org/metamorpho/pull/346 : Wrong rounding in `maxDeposit()`
- https://github.com/morpho-org/metamorpho/pull/342 : Fix : Ensure vaults decimals to be equal to 18 when it's possible.
- https://github.com/morpho-org/metamorpho/pull/359 : Fix : Add comments on scenarios that can prevent fee from accruals.
- https://github.com/morpho-org/metamorpho/pull/365
- https://github.com/morpho-org/metamorpho/pull/358 : It was possible to withdraw from not enable markets with the `reallocate` function.
- https://github.com/morpho-org/metamorpho/pull/366
- https://github.com/morpho-org/metamorpho/pull/363
- https://github.com/morpho-org/metamorpho/pull/354 : Fix : Complete natspec of `submitMarketRemoval`(add comments about the purpose of this mechanism).
- https://github.com/morpho-org/metamorpho/pull/353
- https://github.com/morpho-org/metamorpho/pull/338 : Fee accrued from suppliers assets when removed markets were re-enabled.
- https://github.com/morpho-org/metamorpho/pull/369
- https://github.com/morpho-org/metamorpho/pull/368 : Fix : Add comments explaining that the returned value may be wrong for the `maxDeposit()` and `maxMint()` functions because of duplicate markets in the supply queue.
- https://github.com/morpho-org/metamorpho/pull/370
- https://github.com/morpho-org/metamorpho/pull/371 : Acknowledge that the reallocating funds might give unexpected results


README.md
```diff
@@ -28,8 +28,8 @@ Those rewards can be transferred to the `skimRecipient`.
 The vault's owner has the choice to distribute back these rewards to vault depositors however they want.
 For more information about this use case, see the [Rewards](#rewards) section.
 
-All actions that may be against users' interests (e.g. enabling a market with a high exposure, increasing the fee) are subject to a timelock of minimum 12 hours.
-If set, the `guardian` can revoke the action during the timelock except for the fee increase.
+All actions that may be against users' interests (e.g. enabling a market with a high exposure) are subject to a timelock of minimum 24 hours.
+If set, the `guardian` can revoke the action during the timelock.
```
StErMi, line 32: The `guardian` (or the `owner` itself because he's also a guardian) can always remove the action. `revokePending*` will never revert.

README.md
```diff
@@ -102,9 +104,9 @@ To achieve this, it is advised to allocate "idle" funds to any market on Morpho
 
 - The vault's asset as loan token.
 - No collateral token (`address(0)`).
-- An arbitrary IRM.
+- An IRM that does not trigger a revert (`address(0)`).
 - An arbitrary oracle (`address(0)`).
-- An arbitrary LLTV.
+- An arbitrary LLTV (`0`).
```
StErMi, line 109: Consider removing the values in parentheses and adding an example of suggested config to create a proper idle market. Explain properly which are the values that should be used and why it's important to pick those values. For example, you must set the collateral token to `address(0)` to prevent users from borrowing the supplied liquidity (and so on for each parameter)

src/MetaMorpho.sol
```diff
@@ -375,7 +375,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 withdrawn = supplyAssets.zeroFloorSub(allocation.assets);
 
             if (withdrawn > 0) {
-                if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
+                if (!config[id].enabled) revert ErrorsLib.MarketNotEnabled(id);
```
StErMi, line 378: Should be possible to withdraw from a market that has the pending removal on timelock?

src/interfaces/IMetaMorpho.sol
```diff
@@ -70,36 +70,49 @@ interface IMetaMorphoBase {
     function withdrawQueueLength() external view returns (uint256);
 
     /// @notice Stores the total assets managed by this vault when the fee was last accrued.
-    /// @dev May be a little off `totalAssets()` after each interaction, due to some roundings.
+    /// @dev May be greater than `totalAssets()` due to removal of markets with non-zero supply or socialized bad debt.
```
StErMi, line 73: Consider that this is only temporary until a "sync" operation is triggered

src/interfaces/IMetaMorpho.sol
```diff
@@ -70,36 +70,49 @@ interface IMetaMorphoBase {
...
     function revokePendingCap(Id id) external;
 
-    /// @notice Submits a forced market removal from the vault, potentially losing all funds supplied to the market.
-    /// @dev Warning: Submitting a forced removal will overwrite the timestamp at which the market will be removable.
-    function submitMarketRemoval(Id id) external;
+    /// @notice Submits a forced market removal from the vault, eventually losing all funds supplied to the market.
+    /// @notice Funds can be recovered by enabling this market again and withdrawing from it (using `reallocate`),
+    /// but funds will be distributed pro-rata to the shares at the time of withdrawal, not at the time of removal.
+    /// @notice This forced removal is expected to be used as an emergency process in case a market constantly reverts.
+    /// To softly remove a sane market, the curator role is expected to bundle a reallocation that empties the market
```
StErMi, line 107: Consider also suggesting setting the cap to `0` to prevent any additional supply to a market that needs to be removed via the "normal" way. The market in this case can be removed via `updateWithdrawQueue` only the supply balance is equal to zero.

StErMi: Probably it's useless because you can be front-runned by a donation on behalf of MM. But it's still a good practice, I think

src/MetaMorpho.sol
MerlinEgalite: Yes, users should be able to exit before it's removed. It's required to keep the vault noncustodial.

src/interfaces/IMetaMorpho.sol
MerlinEgalite: you mean we should make it clearer?

src/interfaces/IMetaMorpho.sol
MerlinEgalite: If `submitMarketRemoval` this is required anyway to set the cap to 0 so I think it's not worth adding this comment

README.md
MerlinEgalite: Hmm the values in parenthesis are the ones that should be set to create a canonic idle market. Do you have something else in mind?

src/MetaMorpho.sol
StErMi: Is this comment valid also for the `reallocate` function?

src/interfaces/IMetaMorpho.sol
StErMi: yes

src/interfaces/IMetaMorpho.sol
StErMi: As I suggested in another comment, I think that the best option will be to do a separate guide to list all the these "emergency" scenarios and what actions you need to perform to handle them. Also, having a utility contract that batch them (and have conditions inside based on the state of the pending proposals) will be nice to have.

README.md
StErMi: I think that being explicit with a proper example (explaining why you need to do it like that and specifying those specific values) would make things easy. Being explicit and restrictive is always better.

Also, explaining why to pick those values (and the unwanted side effects of not choosing them) adds value. For example, `address(0)` for the collateral is needed to prevent MB users to borrowing the idle liquidity.

I think that it's fair to think that people that will use MM are less deep into the knowledge and code as you (the owner and developer) are.

src/MetaMorpho.sol
MerlinEgalite: Yes it is, for reallocate the goal is to retrieve funds from this market before it's removed

src/interfaces/IMetaMorpho.sol
MerlinEgalite: Just created an issue for that #375

src/interfaces/IMetaMorpho.sol
MerlinEgalite: I think this is what the line below is mentioning but maybe it's not clear enough

Issue #219, opened 2023-10-18
[protocol review] Check enabled market in `reallocate` withdraw loop


QGarchery:
The `reallocate` does not check that the withdrawn markets are in the `withdrawQueue`. This means that assets could come from unknown parties (legal problem ?) not whitelisted by the curator, potentially after someone has supplied on behalf of the vault on those markets. It also means that the idle amount could be inflated with amounts that were not supplied on the vault

QGarchery:
Re-opening this because I don't understand why we would want that behavior. This is in the same vein as https://github.com/morpho-org/metamorpho/issues/218: basically it's the last safety measure to ensure that allocators cannot interact with markets that have not been previously approved by the curator. The check should only be about checking that the `withdrawRank` of the market is not 0

MerlinEgalite:
It seems there's [no legal reason](https://morpholabs.slack.com/archives/C02NE4SAE4B/p1697723365712959) to do it and it does not seem to be a requirement either

QGarchery:
To provide a little bit more context on why this was closed as not planned: now that there is a way to "force remove" a market (maybe because it is broken), it becomes convenient to be able to rescue funds without having to wait for the timelock. Additionally, putting the cap above 0 on a market that is scheduled for removal is awkward because it potentially  leaves a window for users to supply on that market before it is actually removed

PR #358, opened 2023-12-22
fix(reallocate): prevent withdrawing from market not enabled

Rubilmax:
Fixes https://cantina.xyz/code/8409a0ce-6c21-4cc9-8ef2-bd77ce7425af/findings/131b79f3-83fc-4674-a222-c5c4dc7d0539

But most importantly, fixes the fact that a withdrawn donation gets taxed from the vault manager

1. If the withdrawn donation corresponds to some liquidity the vault owned formerly, this liquidity gets taxed for no reason (for example: the curator bundles `updateWithdrawQueue` then `reallocate` instead of `reallocate` then `updateWithdrawQueue`)
2. The entrypoint `reallocate` now reverts as soon as at least 1 market of the withdraw queue is broken. The curator no longer can reallocate liquidity across working markets as soon as 1 market is broken: they are forced to `submitMarketRemoval` and cannot reallocate for the duration of the timelock
3. it increases the gas cost of `reallocate` in hardhat tests by 55%. This is because in hardhat tests, we most of the time reallocate liquidity between 1-3 markets while the withdraw queue contains 5 markets

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -294,7 +294,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     /// @inheritdoc IMetaMorphoBase
     function submitMarketRemoval(Id id) external onlyCuratorRole {
         if (config[id].removableAt != 0) revert ErrorsLib.AlreadySet();
-        if (!config[id].enabled) revert ErrorsLib.MarketNotEnabled();
+        if (!config[id].enabled) revert ErrorsLib.MarketNotEnabled(id);
 
         _setCap(id, 0);
 
@@ -365,9 +365,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @inheritdoc IMetaMorphoBase
     function reallocate(MarketAllocation[] calldata allocations) external onlyAllocatorRole {
-        // Accrue fee to avoid taking a fee on withdrawn donations.
-        _accrueFee();
-
         uint256 totalSupplied;
         uint256 totalWithdrawn;
         for (uint256 i; i < allocations.length; ++i) {
@@ -378,7 +375,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 withdrawn = supplyAssets.zeroFloorSub(allocation.assets);
 
             if (withdrawn > 0) {
-                if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
+                if (!config[id].enabled) revert ErrorsLib.MarketNotEnabled(id);
 
                 // Guarantees that unknown frontrunning donations can be withdrawn, in order to disable a market.
                 uint256 shares;
@@ -416,9 +413,6 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         }
 
         if (totalWithdrawn != totalSupplied) revert ErrorsLib.InconsistentReallocation();
-
-        // Update `lastTotalAssets` to avoid taking a fee on withdrawn donations.
-        _updateLastTotalAssets(totalAssets());
     }
 
     /* REVOKE FUNCTIONS */
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -229,7 +229,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newFee > ConstantsLib.MAX_FEE) revert ErrorsLib.MaxFeeExceeded();
         if (newFee != 0 && feeRecipient == address(0)) revert ErrorsLib.ZeroFeeRecipient();
 
-        // Accrue interest using the previous fee set before changing it.
+        // Accrue fee using the previous fee set before changing it.
         _updateLastTotalAssets(_accrueFee());
 
         // Safe "unchecked" cast because newFee <= MAX_FEE.
@@ -243,7 +243,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
         if (newFeeRecipient == feeRecipient) revert ErrorsLib.AlreadySet();
         if (newFeeRecipient == address(0) && fee != 0) revert ErrorsLib.ZeroFeeRecipient();
 
-        // Accrue interest to the previous fee recipient set before changing it.
+        // Accrue fee to the previous fee recipient set before changing it.
         _updateLastTotalAssets(_accrueFee());
 
         feeRecipient = newFeeRecipient;
@@ -365,6 +365,9 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
     /// @inheritdoc IMetaMorphoBase
     function reallocate(MarketAllocation[] calldata allocations) external onlyAllocatorRole {
+        // Accrue fee to avoid taking a fee on withdrawn donations.
+        _accrueFee();
+
         uint256 totalSupplied;
         uint256 totalWithdrawn;
         for (uint256 i; i < allocations.length; ++i) {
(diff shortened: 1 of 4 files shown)


Rubilmax:
2 is what worries me the most: at worst, a vault manager cannot reallocate liquidity during the duration of their timelock
But it may be fine

MerlinEgalite:
I don't think taxing donations is a big deal. It's bonus for everyone anyway. 

> But it may be fine
I don't think it's fine. Like if one market is frozen it could be for any reason including reason that could affect similar assets. The allocator would not be able to rebalance the portfolio to reduce risk exposure.

The increase of the gas cost is worrying. IMO reallocation should be as cheap as possible.

Rubilmax:
> It's bonus for everyone anyway.

Not in all cases, see point 1 + in case of a forced market removal that no longer reverts for some reason

MerlinEgalite:
> Not in all cases, see point 1 + in case of a forced market removal that no longer reverts for some reason

But this is fixed by https://github.com/morpho-org/metamorpho/pull/338 right?

Jean-Grimal:
> fixes the fact that a withdrawn donation gets taxed from the vault manager

Romain is right. https://github.com/morpho-org/metamorpho/pull/338 allows to recover (without being "taxed" by the feeRecipient) previously lost supply on reverting markets only when these markets are enabled again. But without enabling these markets it's possible (if they stop reverting), it's possible to withdraw from these market (to recover the supply) with `reallocate`. And in this case, if we don't accrue interest at the start of the function, fee will accrue from this supply.

MerlinEgalite:
Ok right I get it now, sorry

MerlinEgalite:
what's the cost of reallocate @Rubilmax ?

Rubilmax:
> what's the cost of reallocate @Rubilmax ?

according to hardhat tests, 360k gas before this change, 560k gas now (on avg)

MerlinEgalite (changes requested):
This must be tested I think

MerlinEgalite (changes requested):
a (simpler?) solution would be to prevent withdrawing from a market that is not listed? It removes the possibility to withdraw donations but maybe the spec is clearer?

Jean-Grimal (dismissed):

src/MetaMorpho.sol
```diff
@@ -375,7 +375,7 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
             uint256 withdrawn = supplyAssets.zeroFloorSub(allocation.assets);
 
             if (withdrawn > 0) {
-                if (allocation.marketParams.loanToken != asset()) revert ErrorsLib.InconsistentAsset(id);
+                if (!config[id].enabled) revert ErrorsLib.MarketNotEnabled(id);
```
MerlinEgalite, line 378: should we move it outside of the `if`?

test/forge/ReallocateWithdrawTest.sol
```diff
@@ -59,7 +59,7 @@ contract ReallocateWithdrawTest is IntegrationTest {
         assertEq(_idle(), INITIAL_DEPOSIT, "idle");
     }
 
-    function testReallocateWithdrawInconsistentAsset() public {
+    function testReallocateWithdrawMarketNotEnabled() public {
         ERC20Mock loanToken2 = new ERC20Mock("loan2", "B2");
         allMarkets[0].loanToken = address(loanToken2);
```
MerlinEgalite, line 64: might need to be updated as well and/or fuzz

src/MetaMorpho.sol
Rubilmax: No because the supply cap is checked upon supply and `supplyCap > 0 => enabled == true`

test/forge/ReallocateWithdrawTest.sol
Rubilmax: Why would you want to update it?
And fuzzing it is not an option, so we simply deploy a second token (or show me the way!)

src/MetaMorpho.sol
MerlinEgalite: yes ofc, but we could have reverted earlier, even if i know we don't really care

test/forge/ReallocateWithdrawTest.sol
MerlinEgalite: Markets with an inconsistent asset is a subset of markets that are not enabled, so i'd propose to test the function on a larger set like fuzzing the market params. But this is not really important

src/MetaMorpho.sol
Rubilmax: In the case of the supply we don't want to revert, we want to continue

test/forge/ReallocateWithdrawTest.sol
Rubilmax: Ah but you're right there's no call involved so we could fuzz

Rubilmax: But we need a real loan token to supply on behalf of the vault

src/MetaMorpho.sol
MerlinEgalite: indeed

MerlinEgalite (approved):

MathisGD (commented):
I'm not sure to agree with this fix. 
- if the manager enables the market, the donation/previously loosed funds will be slashed right ?
- why shouldn't a donation profit to the fee too ?

QGarchery:
> if the manager enables the market, the donation/previously loosed funds will be slashed right ?

(in english loosed -> lost, and also loose -> lose)
If by "slashed" you mean that it will be subject to the MM fee, then yes. 

> why shouldn't a donation profit to the fee too ?

I don't think it's a problem that a donation can be subject to the MM fee. In effect, donations to enabled markets are subject to the fee, even with the changes of this PR. We can imagine vaults making profit in other ways, and this can be cleanly accounted for with a donation.
What's more problematic is if funds of forced-removed markets are subject to the fee. This is because in this case the fee is taken on the principal, and not on the interest.
Moreover, this change is a safety measure that was already requested in #219, and marked as not planned because not doing it allowed to recover funds  more cleanly from previously forced-removed markets. Indeed, enabling such markets again previously added them in the supply queue, so there would be some weird time where users would be able to supply on broken markets. Markets are not automatically added to the supply queue anymore (since #351), so it enables doing this change without problems

src/libraries/ErrorsLib.sol
```diff
@@ -64,7 +64,7 @@ library ErrorsLib {
     error MarketNotCreated();
 
     /// @notice Thrown when submitting a non previously enabled market for removal.
-    error MarketNotEnabled();
+    error MarketNotEnabled(Id id);
```
QGarchery, line 67: Let's change also the natspec to
```
/// @notice Thrown when interacting with a non previously enabled market `id`.
```
Notice 2 changes: 
- added `id` to the natspec
- the error is not just emitted on forced market removals

QGarchery (approved):

adhusson (approved):

MerlinEgalite (approved):

adhusson (approved):

Rubilmax (commented):
LGTM!

MathisGD (approved):

StErMi:
> The entrypoint reallocate now reverts as soon as at least 1 market of the withdraw queue is broken. The curator no longer can reallocate liquidity across working markets as soon as 1 market is broken: they are forced to submitMarketRemoval and cannot reallocate for the duration of the timelock

@Rubilmax would you mind explaining this? I don't see how the PR has changed this behavior, compared to before

MerlinEgalite:
> > The entrypoint reallocate now reverts as soon as at least 1 market of the withdraw queue is broken. The curator no longer can reallocate liquidity across working markets as soon as 1 market is broken: they are forced to submitMarketRemoval and cannot reallocate for the duration of the timelock
> 
> @Rubilmax would you mind explaining this? I don't see how the PR has changed this behavior, compared to before

This is indeed no longer true (it has been removed in the second commit of this PR).

StErMi:
If you can update the description of the PR with the logic that you want to achieve with the changes, it would allow me to verify that the code is correct and has no side effects.

In general:

1) what it should do
2) what it should not change

MerlinEgalite:
> If you can update the description of the PR with the logic that you want to achieve with the changes, it would allow me to verify that the code is correct and has no side effects.
> 
> In general:
> 
> 1. what it should do
> 2. what it should not change

I updated the description

PR #363, opened 2023-12-27
Simplify comment about removed markets

### LINKED ISSUE #362: misleading comment
https://github.com/morpho-org/metamorpho/blob/0b63c348445fea57532e3dd5d4962961d770d5a6/src/interfaces/IMetaMorpho.sol#L143

what do we mean here?

QGarchery:
Refactor because it is already explained below the different cases where the market can be removed (also the first comment was missing a case)

fixes #362

### DIFF
--- a/src/interfaces/IMetaMorpho.sol
+++ b/src/interfaces/IMetaMorpho.sol
@@ -137,9 +137,8 @@ interface IMetaMorphoBase {
     /// increase the cost of depositing to the vault.
     function setSupplyQueue(Id[] calldata newSupplyQueue) external;
 
-    /// @notice Sets the withdraw queue as a permutation of the previous one, although markets with both zero cap and
-    /// zero vault's supply can be removed from the permutation.
-    /// @notice This is the only entry point to disable a market.
+    /// @notice Sets the withdraw queue as a permutation of the previous one, although some markets can be removed.
+    /// @notice This is the only entry point to remove a market.
     /// @notice Removing a market requires the vault to have 0 supply on it, or to have previously submitted a removal
     /// for this market (with the function `submitMarketRemoval`).
     /// @notice Warning: Anyone can supply on behalf of the vault so the call to `updateWithdrawQueue` that expects a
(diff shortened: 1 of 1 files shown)


Jean-Grimal (commented):
Not sure about this one

src/interfaces/IMetaMorpho.sol
```diff
@@ -137,9 +137,8 @@ interface IMetaMorphoBase {
     /// increase the cost of depositing to the vault.
     function setSupplyQueue(Id[] calldata newSupplyQueue) external;
 
-    /// @notice Sets the withdraw queue as a permutation of the previous one, although markets with both zero cap and
-    /// zero vault's supply can be removed from the permutation.
-    /// @notice This is the only entry point to disable a market.
+    /// @notice Sets the withdraw queue as a permutation of the previous one, although some markets can be removed.
+    /// @notice This is the only entry point to remove a market.
```
adhusson, line 141:
```suggestion
    /// @notice Updates the withdraw queue. No market can be added, some can be removed (see below).
    /// @notice This is the only entry point to remove a market.
```

MerlinEgalite: This version is simple and clear I like it

Rubilmax:
```suggestion
    /// @notice Sets the withdraw queue as a permutation of the previous one, although some markets can be removed.
    /// @notice This is the only entry point to remove a market from the withdraw queue.
```

(optional)

Rubilmax (approved):

src/interfaces/IMetaMorpho.sol
MathisGD, line 141: I still find this comment misleading (#362) even if I understand what it means

MerlinEgalite: What's your prop?

MathisGD: remove it

MathisGD:
```suggestion
```

MathisGD, line 141:
```suggestion
    /// @notice Updates the withdraw queue. Some markets can be removed, but no market can be added.
```

Rubilmax (approved):

MerlinEgalite (approved):

MathisGD (approved):

Jean-Grimal (approved):

PR #284, opened 2023-11-01
Check duplicates in supply queue 109

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/109

PR #367, opened 2023-12-27
supply queue cannot have duplicates

adhusson:
Supply queue is sorted and checked for duplicates on `setSupplyQueue`.

- Sort chosen to be easy to audit/prove correct, not for performance. But at these sizes and for an admin function I don't think it matters.
- The original supply queue is logged, not the sorted one

### DIFF
--- a/src/MetaMorpho.sol
+++ b/src/MetaMorpho.sol
@@ -318,8 +318,27 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
 
         if (length > ConstantsLib.MAX_QUEUE_LENGTH) revert ErrorsLib.MaxQueueLengthExceeded();
 
+        Id[] memory sorted = new Id[](length);
         for (uint256 i; i < length; ++i) {
             if (config[newSupplyQueue[i]].cap == 0) revert ErrorsLib.UnauthorizedMarket(newSupplyQueue[i]);
+            sorted[i] = newSupplyQueue[i];
+        }
+
+        // Gnome sort. O(n) if already sorted.
+        uint256 j;
+        while (j < length) {
+            if (j == 0 || Id.unwrap(sorted[j]) >= Id.unwrap(sorted[j - 1])) {
+                j++;
+            } else {
+                (sorted[j], sorted[j - 1]) = (sorted[j - 1], sorted[j]);
+                j--;
+            }
+        }
+
+        Id prev;
+        for (uint256 i; i < length; ++i) {
+            if (Id.unwrap(prev) == Id.unwrap(sorted[i])) revert ErrorsLib.DuplicateMarket(sorted[i]);
+            prev = sorted[i];
         }
 
         supplyQueue = newSupplyQueue;
--- a/test/forge/MarketTest.sol
+++ b/test/forge/MarketTest.sol
@@ -93,6 +93,17 @@ contract MarketTest is IntegrationTest {
         assertEq(Id.unwrap(vault.supplyQueue(1)), Id.unwrap(allMarkets[2].id()));
     }
 
+    function testSetSupplyQueueWithDuplicate() public {
+        Id[] memory supplyQueue = new Id[](3);
+        supplyQueue[0] = allMarkets[1].id();
+        supplyQueue[1] = allMarkets[2].id();
+        supplyQueue[2] = allMarkets[1].id();
+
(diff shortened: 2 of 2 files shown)


Rubilmax:
This was formerly suggested in https://github.com/morpho-org/metamorpho/pull/284

adhusson:
I looked at the gas with 30 markets in this PR, it's +190k if given without sorting, and +30k if given already sorted.

Thanks for the pointer I missed #284 and https://github.com/cantinasec/review-morpho-blue-1/issues/109. To me the main reason is to make sure the values returned by `maxDeposit` and `maxMint` are correct, but it may not be worth the complexity.

Rubilmax:
> I looked at the gas with 30 markets in this PR, it's +190k if given without sorting, and +30k if given already sorted.
> 
> Thanks for the pointer I missed #284 and [cantinasec/review-morpho-blue-1#109](https://github.com/cantinasec/review-morpho-blue-1/issues/109). To me the main reason is to make sure the values returned by `maxDeposit` and `maxMint` are correct, but it may not be worth the complexity.

These numbers are low enough to me for an admin function so I believe it is worth being merged

src/MetaMorpho.sol
```diff
@@ -318,8 +318,27 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
+        Id[] memory sorted = new Id[](length);
         for (uint256 i; i < length; ++i) {
             if (config[newSupplyQueue[i]].cap == 0) revert ErrorsLib.UnauthorizedMarket(newSupplyQueue[i]);
+            sorted[i] = newSupplyQueue[i];
+        }
+
+        // Gnome sort. O(n) if already sorted.
+        uint256 j;
+        while (j < length) {
+            if (j == 0 || Id.unwrap(sorted[j]) >= Id.unwrap(sorted[j - 1])) {
```
Rubilmax, line 330: I feel like `sorted[j]` & `sorted[j-1]` could be stored as variables not only to be cheaper (max hundred gas) but also to be clearer

Rubilmax: Like `curr = sorted[j]`, `prev = sorted[j-1]`

adhusson: I think you mean this but let me know if I misunderstood

```solidity
        uint256 j;
        Id cur;
        Id prev;
        while (j < length) {
            if (j==0) {
                j++;
            } else {
                cur = sorted[j];
                prev = sorted[j-1];
                if (Id.unwrap(cur) >= Id.unwrap(prev)) {
                    j++;
                } else {
                    (sorted[j], sorted[j - 1]) = (prev, cur);
                    j--;
                }
            }
        }

        prev = Id.wrap(0);
        for (uint256 i; i < length; ++i) {
            if (Id.unwrap(prev) == Id.unwrap(sorted[i])) revert ErrorsLib.DuplicateMarket(sorted[i]);
            prev = sorted[i];
        }
``` 

it's -23k gas (out of ~350k)

Rubilmax: Indeed! Not sure it's better in the end

The last for loop is maybe too much too, simply naming variables in the while loop could be enough

adhusson: to clarify, you're saying it may be worth moving duplicate detection inside the while loop. I think it's correct because:

- Some pair of equal elements in the array must end up with both elements adjacent to each other. Take the leftmost such pair.
- If the pair was already adjacent before sorting and never moved, since the sort looks at all initially adjacent pairs once, they will be detected
- If the last move that made the pair adjacent moved the right element of the pair into position, the next comparison is between the elements of the pair (`j--`), so they will be detected
- If the last move that made the pair adjacent moved the left element of the pair into position, the next comparison will go left (`j--`) but the loop will not end until `j == length` and so the elements will compared and detected.

That being said it may not be worth it in gas

MathisGD:
is there a consensus on if we need this pr or not? I feel that we can just delegate to the risk manager to not have duplicates, and have maxDeposit and maxMint rely on this trust. wdyt?

MathisGD:
did we think about having a mapping `isInSupplyQueue` ? It increases the storage size but costs only to the manager, and prevents having this O(n**2) check

MerlinEgalite:
Just had a call with @MathisGD and @Jean-Grimal. We think this PR adds too much complexity while it should be the responsibility of the allocator not to let duplicate markets in the `supplyQueue`. We prefer just adding a comment see #368.

PR #371, opened 2023-12-28
acknowledge non idempotence of `updateWithdrawQueue`

### LINKED ISSUE #316: `updateWithdrawQueue` can be front-runned by other allocators so the end result depends on the order of txs
# Context

`updateWithdrawQueue` expects `indexes` as input which is a mapping from the new expected index in the withdraw queue to the previous index in the withdraw queue

The issue is that "the previous index in the withdraw queue" depends on the state of the blockchain before the CALL, which may not be as expected, especially when multiple allocators submit a new withdraw queue order

For example:
0. Withdraw queue is [id0, id1, id2]
1. Allocator 1 submits [1, 2, 0]
2. Allocator 2 submits [1, 2, 0]
3. Allocator 1's tx gets included: withdraw queue is [id1, id2, id0]
4. Allocator 2's tx gets included: withdraw queue is [id2, id0, id1]

# Recommendation

The issue is similar to slippage, so we can build a slippage protection mechanism: expect each allocator to input `(uint256 index, Id id)[] indexes` and validate that `id == withdrawQueue[index]` for each element in it, otherwise revert

In the previous example, it would correspond to: Allocator 2's tx reverts are doesn't get included in the blockchain

MathisGD:
Fixes #316 (by acknowledging it)

discussion: https://morpholabs.slack.com/archives/C02N7CSP91C/p1703775229246109

### DIFF
--- a/src/interfaces/IMetaMorpho.sol
+++ b/src/interfaces/IMetaMorpho.sol
@@ -159,6 +159,7 @@ interface IMetaMorphoBase {
     /// market to be empty can be griefed by a front-run. To circumvent this, the allocator can simply bundle a
     /// reallocation that withdraws max from this market with a call to `updateWithdrawQueue`.
     /// @dev Warning: Removing a market with supply will decrease the fee accrued until the next deposit/withdrawal.
+    /// @dev Warning: updateWithdrawQueue is not idempotent. Submitting twice the same tx will change the queue twice.
     /// @param indexes The indexes of each market in the previous withdraw queue, in the new withdraw queue's order.
     function updateWithdrawQueue(uint256[] calldata indexes) external;
 


MerlinEgalite (approved):

Jean-Grimal (approved):

src/interfaces/IMetaMorpho.sol
```diff
@@ -159,6 +159,7 @@ interface IMetaMorphoBase {
     /// market to be empty can be griefed by a front-run. To circumvent this, the allocator can simply bundle a
     /// reallocation that withdraws max from this market with a call to `updateWithdrawQueue`.
     /// @dev Warning: Removing a market with supply will decrease the fee accrued until the next deposit/withdrawal.
+    /// @dev Warning: updateWithdrawQueue is not idempotent. Submitting twice the same tx will change the queue twice.
```
StErMi, line 162: Nitpick: updateWithdrawQueue here is not wrapped by ` but in the above comments it is.

MerlinEgalite: nitpicking goes brrrrr

StErMi: > nitpicking goes brrrrr

it's a trap!

PR #378, opened 2024-01-03
build: update optimizer

MerlinEgalite:
Removes the optimizer runs because the factory is too heavy to be deployed with optimizations run

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -7,6 +7,7 @@ fs_permissions = [
     { access = "read", path = "./config/"},
     { access = "read", path = "./lib/morpho-blue/out/"}
 ]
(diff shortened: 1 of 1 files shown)


foundry.toml
QGarchery, line 1: Can you also change the last line of this file to put the correct link:
https://github.com/foundry-rs/foundry/tree/master/crates/config

QGarchery (approved):

MathisGD (approved):

Rubilmax:
What was the rationale behind switching back to the default optimization runs of 200?

MerlinEgalite:
> What was the rationale behind switching back to the default optimization runs of 200?

If you try even 1_000 you exceed the contract size limit for the factory...

PR #380, opened 2024-01-05
Add emergency section in the README

### LINKED ISSUE #375: Create emergency section in README with step actions
As I suggested in another comment, I think that the best option will be to do a separate guide to list all the these "emergency" scenarios and what actions you need to perform to handle them. Also, having a utility contract that batch them (and have conditions inside based on the state of the pending proposals) will be nice to have.

_Originally posted by @StErMi in https://github.com/morpho-org/metamorpho/pull/357#discussion_r1438833324_

Jean-Grimal:
Fixes #375 

Should I add subsections about rogue guardian and owner takeover ?
I haven't done it yet because it require to deploy a new vault and to migrate positions to this new vault, and I was not sure such operations should be explained/described in this file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,6 @@ It is advised to use these canonical configurations for "idle" markets:
 - `oracle`: `address(0)` (not necessary since no funds will be borrowed on this market)
 - `lltv`: `0` (not necessary since no funds will be borrowed on this market)
 
-
 Note that to allocate funds to this idle market, it is first required to enable its cap on MetaMorpho.
 Enabling an infinite cap (`type(uint184).max`) will always allow users to deposit on the vault.
 
@@ -137,6 +136,44 @@ Below is a typical example of how this use case would take place:
   Note: the amount of rewards transferred corresponds to the vault's balance of reward asset.
 
 - Compute the new root for the vault’s rewards distributor, submit it, wait for the timelock (if any), accept the root, and let vault depositors claim their rewards according to the vault manager’s rewards re-distribution strategy.
+
+## Emergency
+
+### An enabled market risk is now considered unsafe
+
+If the owner/curator consider an enabled market unsafe (e.g., risk too high) and want to disable this market, he should follow these steps :
+
+- Set the cap of the market to 0 with the `submitCap` function.
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Withdraw all the supply of this market with the `reallocate` function. If there is not enough liquidity on the market, the market should be put at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
+- Once all the supply has been removed from the market, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### An enabled market reverts
+
+If an enabled market starts reverting, it makes a lot of the vault functions revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
+
+- Set the cap of the market to 0 with the `submitCap` function.
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Submit a removal of the market with the `submitMarketRemoval` function.
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### Curator takeover
+
+If the curator starts to submits positive caps for unsafe markets that are not in line with the vault risk strategy, the owner of the vault should :
+
+- Set a new curator with the `setCurator` function.
+- Revoke the pending caps submitted by the guardian (this can also be done by the guardian or the new curator).
+- If the curator had the time to accept a cap (because `timelock` has elapsed before the guardian or the owner had time to act), the owner (or the new curator) must disable the unsafe market (see the "An enabled market risk is now considered unsafe" section above).
+
+### Allocator takeover
+
+If one of the allocators starts setting withdraw queue and/or supply queue that are not in line with the vault risk strategy, or maliciously reallocating the funds, the owner of the vault should :
+
+- Deprive the malicious allocator from his privileges with the `setIsAllocator` function.
+- Reallocate the funds in a way consistent with the vault risk strategy with the `reallocate` function (this can also be done by the currator or the other allocators).
+- Set a new withdraw queue that is in line with the vault risk strategy with the `updateWithdrawQueue` function (this can also be done by the currator or the other allocators).
+- Set a new supply queue that is in line with the vault risk strategy with the `setSupplyQueue` function (this can also be done by the currator or the other allocators).
 
 ## Getting Started
 


README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
 
 - Compute the new root for the vault’s rewards distributor, submit it, wait for the timelock (if any), accept the root, and let vault depositors claim their rewards according to the vault manager’s rewards re-distribution strategy.
 
+## Emergency
+
+### An enabled market risk is now considered unsafe
```
QGarchery, line 142:
```suggestion
### An enabled market is now considered unsafe
```

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+
+If the owner/curator consider an enabled market unsafe (e.g., risk too high) and want to disable this market, he should follow these steps :
```
QGarchery, line 144:
```suggestion
If an enabled market is considered unsafe (e.g., risk too high), the curator/owner may want to disable this market in the following way:
```
[optional]

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+
+- Set the cap of the market to 0 with the `submitCap` function.
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Withdraw all the supply of this market with the `reallocate` function. If there is not enough liquidity on the market, the market should be put at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
```
QGarchery, line 148:
```suggestion
- Withdraw all the supply of this market with the `reallocate` function.
  If there is not enough liquidity on the market, the market should be put at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
```
putting each sentence on its own line is recommended so that changes are easier to track with git

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+- Once all the supply has been removed from the market, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### An enabled market reverts
+
+If an enabled market starts reverting, it makes a lot of the vault functions revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
```
QGarchery, line 153: same

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### Curator takeover
+
+If the curator starts to submits positive caps for unsafe markets that are not in line with the vault risk strategy, the owner of the vault should :
+
+- Set a new curator with the `setCurator` function.
+- Revoke the pending caps submitted by the guardian (this can also be done by the guardian or the new curator).
+- If the curator had the time to accept a cap (because `timelock` has elapsed before the guardian or the owner had time to act), the owner (or the new curator) must disable the unsafe market (see the "An enabled market risk is now considered unsafe" section above).
```
QGarchery, line 167: same

README.md
QGarchery, line 148: Can be reworded to explain that it's not all or nothing: remove the maximum (capped by the liquidity), and then for the rest rely on the withdraw queue

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
 - Compute the new root for the vault’s rewards distributor, submit it, wait for the timelock (if any), accept the root, and let vault depositors claim their rewards according to the vault manager’s rewards re-distribution strategy.
 
+## Emergency
+
+### An enabled market risk is now considered unsafe
+
+If the owner/curator consider an enabled market unsafe (e.g., risk too high) and want to disable this market, he should follow these steps :
+
+- Set the cap of the market to 0 with the `submitCap` function.
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
```
QGarchery, line 147: These 2 lines should be switched, otherwise the submit cap can revert because of this line of code:
https://github.com/morpho-org/metamorpho/blob/2a2b25087821a2f079b12a29224220985b4a80ab/src/MetaMorpho.sol#L279

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+- Withdraw all the supply of this market with the `reallocate` function. If there is not enough liquidity on the market, the market should be put at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
+- Once all the supply has been removed from the market, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### An enabled market reverts
+
+If an enabled market starts reverting, it makes a lot of the vault functions revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
+
+- Set the cap of the market to 0 with the `submitCap` function.
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
```
QGarchery, line 156: same

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+- Submit a removal of the market with the `submitMarketRemoval` function.
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### Curator takeover
+
+If the curator starts to submits positive caps for unsafe markets that are not in line with the vault risk strategy, the owner of the vault should :
```
QGarchery, line 163:
```suggestion
If the curator starts to submit positive caps for unsafe markets that are not in line with the vault risk strategy, the owner of the vault can:
```

README.md
QGarchery, line 167: [optional] You can link sections natively in markdown, see [this resource](https://anvilproject.org/guides/content/creating-links#linking-to-an-internal-page-heading)

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+
+- Set a new curator with the `setCurator` function.
+- Revoke the pending caps submitted by the guardian (this can also be done by the guardian or the new curator).
```
QGarchery, line 166:
```suggestion
- Revoke the pending caps submitted by the curator (this can also be done by the guardian or the new curator).
```

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+- If the curator had the time to accept a cap (because `timelock` has elapsed before the guardian or the owner had time to act), the owner (or the new curator) must disable the unsafe market (see the "An enabled market risk is now considered unsafe" section above).
+
+### Allocator takeover
+
+If one of the allocators starts setting withdraw queue and/or supply queue that are not in line with the vault risk strategy, or maliciously reallocating the funds, the owner of the vault should :
```
QGarchery, line 171: I would not say malicously, because it makes it sound like it's possible for the allocator to move funds as they desire (but it's actually not the case by design: the allocator can only move funds among trusted markets)

README.md
```diff
@@ -138,6 +137,44 @@ Below is a typical example of how this use case would take place:
...
+
+- Deprive the malicious allocator from his privileges with the `setIsAllocator` function.
```
QGarchery, line 173:
```suggestion
- Deprive the faulty allocator from his privileges with the `setIsAllocator` function.
```

README.md
QGarchery, line 153: I think we should be more cautious in how this is phrased: markets are not expected to "start reverting". Also it sounds like if a market reverts at some point it will always revert, whereas maybe the correct solution would be just to wait it out a bit

QGarchery, line 153:
```suggestion
If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
```
[optional]

QGarchery (changes requested):

QGarchery (approved):
LGTM

Two other scenario we could document:
1. the guardian suspects bad behavior from the curator/owner -> revoke everything continuously, explain the situation, and can advocate for removing funds if no solution can be found
2. the allocators can pause the deposits (for example in case of a suspected contract bug involving depositing funds) by setting the supplyQueue to the empty list

Do you thing we should add them @morpho-org/onchain-reviewers ?

README.md
```diff
@@ -138,6 +137,45 @@ Below is a typical example of how this use case would take place:
 
 - Compute the new root for the vault’s rewards distributor, submit it, wait for the timelock (if any), accept the root, and let vault depositors claim their rewards according to the vault manager’s rewards re-distribution strategy.
 
+## Emergency
+
+### An enabled market is now considered unsafe
+
+If an enabled market is considered unsafe (e.g., risk too high), the curator/owner may want to disable this market in the following way:
+
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
```
MerlinEgalite, line 146: maybe we should say that transactions must be batched together via the multicall?

README.md
```diff
@@ -138,6 +137,45 @@ Below is a typical example of how this use case would take place:
...
+- Set the cap of the market to 0 with the `submitCap` function.
+- Withdraw all the supply of this market with the `reallocate` function.
+  If there is not enough liquidity on the market, remove the maximum available liquidity with the `reallocate` function, then put the market at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
+- Once all the supply has been removed from the market, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### An enabled market reverts
+
+If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
```
MerlinEgalite, line 154:
```suggestion
If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps:
```

README.md
```diff
@@ -138,6 +137,45 @@ Below is a typical example of how this use case would take place:
...
+
+If an enabled market is considered unsafe (e.g., risk too high), the curator/owner may want to disable this market in the following way:
+
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Set the cap of the market to 0 with the `submitCap` function.
+- Withdraw all the supply of this market with the `reallocate` function.
+  If there is not enough liquidity on the market, remove the maximum available liquidity with the `reallocate` function, then put the market at the beginning of the withdraw queue (with the `updateWithdrawQueue` function).
+- Once all the supply has been removed from the market, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### An enabled market reverts
```
MerlinEgalite, line 152: I think we should say that in this case users will lose funds

README.md
```diff
@@ -138,6 +137,45 @@ Below is a typical example of how this use case would take place:
...
+
+If the curator starts to submit positive caps for unsafe markets that are not in line with the vault risk strategy, the owner of the vault can:
+
+- Set a new curator with the `setCurator` function.
+- Revoke the pending caps submitted by the curator (this can also be done by the guardian or the new curator).
+- If the curator had the time to accept a cap (because `timelock` has elapsed before the guardian or the owner had time to act), the owner (or the new curator) must disable the unsafe market (see [above](#an-enabled-market-is-now-considered-unsafe)).
+
+### Allocator takeover
+
+If one of the allocators starts setting withdraw queue and/or supply queue that are not in line with the vault risk strategy, or incoherently reallocating the funds, the owner of the vault should :
```
MerlinEgalite, line 172:
```suggestion
If one of the allocators starts setting withdraw queue and/or supply queue that are not in line with the vault risk strategy, or incoherently reallocating the funds, the owner of the vault should:
```

MerlinEgalite:
> LGTM
> 
> Two other scenario we could document:
> 
> 1. the guardian suspects bad behavior from the curator/owner -> revoke everything continuously, explain the situation, and can advocate for removing funds if no solution can be found
> 2. the allocators can pause the deposits (for example in case of a suspected contract bug involving depositing funds) by setting the supplyQueue to the empty list
> 
> Do you thing we should add them @morpho-org/onchain-reviewers ?

Indeed I think it's worth it.

README.md
QGarchery: Maybe not "must", because it sometimes works without the multicall, so something like: "To ensure that submit cap does not revert because of a pending cap, it is recommended to batch the two previous transactions, for example using the `multicall` function of MetaMorpho."

README.md
```diff
@@ -138,6 +137,45 @@ Below is a typical example of how this use case would take place:
...
+
+If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps :
+
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Set the cap of the market to 0 with the `submitCap` function.
+- Submit a removal of the market with the `submitMarketRemoval` function.
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+### Curator takeover
```
Rubilmax, line 162: a meme is worth a thousand words
image

Rubilmax: bringing some fun in PR reviews but i understand if you feel overwhelmed. Feel free to resolve

README.md
adhusson: If done in multiple tx, the item should be "withdraw as much supply as possible with reallocate".

README.md
```diff
@@ -138,6 +137,49 @@ Below is a typical example of how this use case would take place:
...
+If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps:
+
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Set the cap of the market to 0 with the `submitCap` function.
+  To ensure that submit cap does not revert because of a pending cap, it is recommended to batch the two previous transactions, for example using the multicall function of MetaMorpho.
+- Submit a removal of the market with the `submitMarketRemoval` function.
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
+
+Warning : Funds supplied in forced removed markets will be lost, this why only reverting market should be disabled this way (because funds supplied in such markets can be considered lost anyway).
```
QGarchery, line 164:
```suggestion
Warning : Funds supplied in forced removed markets will be lost, this is why only market expect to always revert should be disabled this way (because funds supplied in such markets can be considered lost anyway).
```

adhusson:
> we can't prepare bundles because they are specific to MM addresses and configuration?

(I had default Safe bundle skeletons in mind) After discussion with Romain the best would be foundry script with options. That's enough for EOA allocators, and the resulting tx [will also be convertible into Safe bundles](https://github.com/morpho-labs/create-safe-tx). @MerlinEgalite @Rubilmax

Rubilmax:
> > we can't prepare bundles because they are specific to MM addresses and configuration?
> 
> (I had default Safe bundle skeletons in mind) After discussion with Romain the best would be foundry script with options. That's enough for EOA allocators, and the resulting tx [will also be convertible into Safe bundles](https://github.com/morpho-labs/create-safe-tx). @MerlinEgalite @Rubilmax

moved the discussion to #383!

README.md
```diff
@@ -138,6 +137,49 @@ Below is a typical example of how this use case would take place:
...
+### An enabled market reverts
+
+If an enabled market starts reverting, many of the vault functions would revert as well (because of the call to `totalAssets`). To turn the vault back to an operating state, the market must be forced removed by the owner/curator, who should follow these steps:
+
+- Revoke the pending cap of the market with the `revokePendingCap` function (this can also be done by the guardian).
+- Set the cap of the market to 0 with the `submitCap` function.
+  To ensure that submit cap does not revert because of a pending cap, it is recommended to batch the two previous transactions, for example using the multicall function of MetaMorpho.
+- Submit a removal of the market with the `submitMarketRemoval` function.
+- Wait for the timelock to elapse
+- Once the timelock has elapsed, the market can be removed from the withdraw queue with the `updateWithdrawQueue` function.
```
Rubilmax, line 162: Nitpicking: this list could be numbered, because steps are order-dependent

Rubilmax (dismissed):

README.md
Jean-Grimal: I think it's implicit that the order matters, but I'm okay to make the list numbered.
Wdyt @morpho-org/onchain-reviewers ?

MerlinEgalite: Make things easier, I agree this is better

MerlinEgalite:
ah formatting is broken it seems @Jean-Grimal

PR #381, opened 2024-01-05
chore: add competition

MerlinEgalite:


QGarchery (changes requested):
I don't think it adds much value to add this to the repo:
- it cannot be seen on github because it's too big
- there are plenty of duplicates and useless issues which makes it difficult to see real issues
- real issues do not show what solution has been chosen. It makes it feel like nothing was done for them, which is not the case
- it's so big that I'm not confident there are no bad language for example (especially since there are other languages than english)

MerlinEgalite:
> I don't think it adds much value to add this to the repo:
> 
> * it cannot be seen on github because it's too big
> * there are plenty of duplicates and useless issues which makes it difficult to see real issues
> * real issues do not show what solution has been chosen. It makes it feel like nothing was done for them, which is not the case
> * it's so big that I'm not confident there are no bad language for example (especially since there are other languages than english)

Note that this document does not contain duplicate/

We need a way to expose the findings from precedent audit and security competitions for the upcoming but bounty. 
What alternative do you suggest?

QGarchery (approved):
> We need a way to expose the findings from precedent audit and security competitions for the upcoming but bounty.
What alternative do you suggest?

Adding it is required to start the bug bounty so I don't see a better alternative: every submitted issue must be shared, and it's easier if it's on the same repo

PR #382, opened 2024-01-08
[Certora] Tokens

QGarchery:
This PR checks that MetaMorpho delegates every asset to Morpho. It does so with 3 rules, for balances change in case of: deposit, withdraw and reallocate


certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+}
+
+function summaryIdToMarketParams(MetaMorphoHarness.Id id) returns MetaMorphoHarness.MarketParams {
+    MetaMorphoHarness.MarketParams marketParams;
+    uint256 lastUpdated = Morpho.lastUpdate(id);
+
+    // Safe require because markets in the supply/withdraw queue have positive last update (see LastUpdated.spec).
+    require lastUpdated > 0;
+    // Safe require because it is a verified invariant in Morpho Blue.
+    require lastUpdated > 0 => Morpho.libId(marketParams) == id;
```
MathisGD, line 30: maybe worth being more explicit on the state, and explain more in the comments?
```suggestion
    require lastUpdated > 0;
    // Safe require because lastUpdated > 0 => Morpho.libId(marketParams) == id is a verified invariant in Morpho Blue.
    require Morpho.libId(marketParams) == id;
```

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+    return marketParams;
+}
+
+function summarySupply(MetaMorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns (uint256, uint256) {
+    assert shares == 0;
+    assert onBehalf == currentContract;
+    assert data.length == 0;
+
+    // Safe require because it is a verified invariant.
+    require hasSupplyCapIsEnabled(Morpho.libId(marketParams));
```
MathisGD, line 41: I don't understand completely the sense of a "require" in a summary. It makes total sense for me for an invariant or a rule, we assume more things to verify or prove something, but in a summary where we are describing an execution I don't really get the sense.

What happens for example if you require a non-sense.

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+
+    MetaMorphoHarness.Id id = Morpho.libId(marketParams);
+    uint256 withdrawn = Util.withdrawnAssets(MORPHO(), id, assets, shares);
+
+    bool enabled;
+    _, enabled, _ = config(id);
+    // Safe require because:
+    // - for reallocate this is checked in the code
+    // - for withdraw, it is verified that markets in the withdraw queue are enabled.
+    require enabled;
```
MathisGD, line 63: I don't understand logically why do we need to require this in order to prove what we want to prove. Even if Metamorpho was supplying or withdrawing from non enabled markets, the token accounting would still be ok no?

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+rule depositTokenChange(env e, uint256 assets, address receiver) {
+    address asset = asset();
+    address morpho = MORPHO();
+
+    // Trick to require that all the following addresses are different.
+    require morpho == 0x10;
+    require asset == 0x11;
+    require currentContract == 0x12;
+    require e.msg.sender == 0x13;
+    require receiver == 0x14;
```
MathisGD, line 83: it should not be an issue that the asset and receiver are the same address as the sender, mm, or blue

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+
+    uint256 balanceMorphoBefore = Util.balanceOf(asset, morpho);
+    uint256 balanceMetaMorphoBefore = Util.balanceOf(asset, currentContract);
+    deposit(e, assets, receiver);
+    uint256 balanceMorphoAfter = Util.balanceOf(asset, morpho);
+    uint256 balanceMetaMorphoAfter = Util.balanceOf(asset, currentContract);
+
+    assert assert_uint256(balanceMorphoAfter - balanceMorphoBefore) == assets;
+    assert balanceMetaMorphoAfter == balanceMetaMorphoBefore;
```
MathisGD, line 92:
```suggestion
    uint256 balanceMorphoBefore = Util.balanceOf(asset, morpho);
    uint256 balanceMetaMorphoBefore = Util.balanceOf(asset, currentContract);
    uint256 balanceSenderBefore = Util.balanceOf(asset, e.msg.sender);
    deposit(e, assets, receiver);
    uint256 balanceMorphoAfter = Util.balanceOf(asset, morpho);
    uint256 balanceMetaMorphoAfter = Util.balanceOf(asset, currentContract);
    uint256 balanceSenderAfter = Util.balanceOf(asset, e.msg.sender);

    assert assert_uint256(balanceMorphoAfter - balanceMorphoBefore) == assets;
    assert balanceMetaMorphoAfter == balanceMetaMorphoBefore;
    assert assert_uint256(balanceSenderBefore - balanceSenderAfter) == assets;
```

I imagine that you thought about it. Why doesn't it work?

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+    address asset = asset();
+    address morpho = MORPHO();
+
+    // Trick to require that all the following addresses are different.
+    require morpho == 0x10;
+    require asset == 0x11;
+    require currentContract == 0x12;
+    require e.msg.sender == 0x13;
+    require receiver == 0x14;
+    require owner == 0x15;
```
MathisGD, line 106: what can happen if e.msg.sender (I imagine that it's this causing issues) is the owner ?

certora/specs/Tokens.spec
```diff
@@ -0,0 +1,137 @@
...
+
+    uint256 balanceMorphoBefore = Util.balanceOf(asset, morpho);
+    uint256 balanceMetaMorphoBefore = Util.balanceOf(asset, currentContract);
+    withdraw(e, assets, receiver, owner);
+    uint256 balanceMorphoAfter = Util.balanceOf(asset, morpho);
+    uint256 balanceMetaMorphoAfter = Util.balanceOf(asset, currentContract);
+
+    assert assert_uint256(balanceMorphoBefore - balanceMorphoAfter) == assets;
+    assert balanceMetaMorphoAfter == balanceMetaMorphoBefore;
```
MathisGD, line 115: same question on the sender balance

certora/specs/Tokens.spec
QGarchery: In the ideal case, this would be a `requireInvariant invariantFromMorpho(marketParams)` so we would not have the choice to change the formula inside. This is not possible currently for a technical reason though. But still I would keep it as is to make it closer to a `requireInvariant` which is sound by default (contrary to `require`).

Completely fine to add more comments though, if something else is unclear

certora/specs/Tokens.spec
QGarchery: `require` from summary behaves the same as a regular require in the rule, except that it is as if it was written inside the function call. Usually it's useful to have access to the proper parameters, because inside the function you can have more context.

As for require elsewhere, if you `require false` then you can prove anything you want, which is obviously dangerous

certora/specs/Tokens.spec
QGarchery: We need it in conjunction with the next invariant. When we have both, we know that the market has a consistent asset, which is useful to show that we have proper accounting. Otherwise we could have a transfer of a token A from the user to MM and then a transfer of token B from MM to Morpho

certora/specs/Tokens.spec
QGarchery: Let me try it out, I think you are right and we can remove some of those. I had an issue when the asset was MM, which sounds like a case that would not happen so I did not bother looking too much into it

certora/specs/Tokens.spec
QGarchery: I need to investigate this more

certora/specs/Tokens.spec
MathisGD: wow indeed, nice

certora/specs/Tokens.spec
MathisGD: I get it

certora/specs/Tokens.spec
MathisGD: get it

certora/specs/Tokens.spec
QGarchery: `receiver` removed.

`asset` cannot be equal to MM though, because the fee (on MM) can mint new shares to the fee recipient which would increase his balance. Then we would need to require that the fee recipient is someone else, but it is just more complicated than the current rule

certora/specs/Tokens.spec
QGarchery: Added, but it's actually the receiver's balance which is increasing

certora/specs/Tokens.spec
QGarchery: Added, it seems to work fine actually

MathisGD (approved):

MerlinEgalite (approved):

gabriellavalecia0-cpu:
> This PR checks that MetaMorpho delegates every asset to Morpho. It does so with 3 rules, for balances change in case of: deposit, withdraw and reallocate

PR #387, opened 2024-01-08
[Certora] Consistent pending values

QGarchery:
This PR's goal is to prove that pending values are different than current values. It also proves that there are no "dirty pending values", ie pending values that are not valid (like a 0 timelock) with 0 timestamp


certora/specs/ConsistentState.spec
```diff
@@ -41,3 +45,112 @@ invariant supplyQueueLengthInRange()
...
+function hasNoBadPendingTimelock() returns bool {
+    uint192 pendingValue;
+    uint64 validAt;
+    pendingValue, validAt = pendingTimelock();
+
+    return validAt == 0 <=> pendingValue == 0;
+}
+
+invariant noBadPendingTimelock()
+    hasNoBadPendingTimelock()
```
MathisGD, line 58: why isn't it inlined?

certora/specs/ConsistentState.spec
```diff
@@ -41,3 +45,112 @@ invariant supplyQueueLengthInRange()
...
+        requireInvariant timelockInRange();
+    }
+}
+
+function isGreaterPendingTimelock() returns bool {
+    uint192 pendingValue;
+    pendingValue, _ = pendingTimelock();
+
+    return assert_uint256(pendingValue) != 0 => assert_uint256(pendingValue) != timelock();
+}
```
MathisGD, line 88: - a timelock is pending iff it is set to a lower value right? So it can only be pending it is smaller
- here is is returned pendingValue!=timelock and not pendingValue<timelock
```suggestion
function isGreaterPendingTimelock() returns bool {
    uint192 pendingValue;
    pendingValue, _ = pendingTimelock();

    return assert_uint256(pendingValue) < timelock();
}
```

certora/specs/ConsistentState.spec
```diff
@@ -41,3 +45,112 @@ invariant supplyQueueLengthInRange()
...
+    }
+}
+
+function isGreaterPendingCap(MetaMorphoHarness.Id id) returns bool {
+    uint192 pendingValue;
+    pendingValue, _ = pendingCap(id);
+    uint192 currentValue;
+    currentValue, _, _ = config(id);
+
+    return pendingValue != 0 => assert_uint256(pendingValue) > assert_uint256(currentValue);
```
MathisGD, line 123: also possible
```suggestion
    return validAt != 0 => assert_uint256(pendingValue) > assert_uint256(currentValue);
```

certora/specs/ConsistentState.spec
```diff
@@ -41,3 +45,112 @@ invariant supplyQueueLengthInRange()
 
 invariant withdrawQueueLengthInRange()
     withdrawQueueLength() <= maxQueueLength();
+
+function hasNoBadPendingTimelock() returns bool {
+    uint192 pendingValue;
+    uint64 validAt;
+    pendingValue, validAt = pendingTimelock();
+
+    return validAt == 0 <=> pendingValue == 0;
```
MathisGD, line 54: we could verify only this here
```suggestion
    return validAt == 0 => pendingValue == 0;
```

certora/specs/ConsistentState.spec
```diff
@@ -41,3 +45,112 @@ invariant supplyQueueLengthInRange()
...
+        requireInvariant timelockInRange();
+    }
+}
+
+function hasNoBadPendingCap(MetaMorphoHarness.Id id) returns bool {
+    uint192 pendingValue;
+    uint64 validAt;
+    pendingValue, validAt = pendingCap(id);
+
+    return validAt == 0 <=> pendingValue == 0;
```
MathisGD, line 104: we could verify only this here
```suggestion
    return validAt == 0 => pendingValue == 0;
```

certora/specs/ConsistentState.spec
```diff
@@ -32,7 +36,7 @@ invariant timelockInRange()
     timelock() <= maxTimelock() && timelock() >= minTimelock()
 {
     preserved {
-        require isPendingTimelockInRange();
+        requireInvariant pendingTimelockInRange();
```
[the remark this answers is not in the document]
QGarchery: It does the same thing (so it does not help the prover), but `requireInvariant` is safer than `require` because invariants are always verified. Whereas with `require` you always have to make sure that the formula that is passed (here it's `isPendingTimelockInRange()`) is true: if it's false then the verification will trivially pass, which makes the rule essentially useless

certora/specs/ConsistentState.spec
QGarchery: True, but it does not cost much to add this, so I think we can keep it

certora/specs/ConsistentState.spec
QGarchery: Yes I think it's also true, but I prefer the current way to keep the symmetry with `isDifferentPendingGuardian`

MathisGD: can't you use also validAt there ?

MathisGD (approved):

adhusson (approved):

PR #388, opened 2024-01-08
chore: use toplevel out/ for importing Morpho bytecode

adhusson:
Fixes the following issue : after a fresh checkout, `forge test` fails. The relevant trace portion is:

```
├─ [0] VM::getCode("lib/morpho-blue/out/Morpho.sol/Morpho.json") [staticcall]
    │   └─ ← failed to read from "~/metamorpho/lib/morpho-blue/out/Morpho.sol/Morpho.json": No such file or directory (os error 2)
```

The current fix is to `cd lib/morpho-blue; forge build; cd ../..`, and to do it again after any change to the local `Morpho.sol`.

This PR uses foundry's [multiple compiler version support](https://github.com/foundry-rs/foundry/pull/4) to compile Morpho separately. No more need to build inside the `lib/morpho-blue` dir to get the Morpho bytecode.

Important: Morpho's solidity version will be used, but metamorpho's optimization settings will be used.


src/mocks/MorphoImport.sol
QGarchery, line 1: It seems like hardhat is doing it manually, so we should add the compiler version in the config file too :/

src/mocks/MorphoImport.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.0.0;
```
QGarchery, line 2:
```suggestion
pragma solidity 0.8.19;
```
[optional]

src/mocks/MorphoImport.sol
```diff
@@ -0,0 +1,7 @@
...
+// Force foundry to compile Morpho Blue even though it's not imported by Metamorpho or by the tests
+// Morpho Blue will be compiled with its own solidity version
+// The resulting bytecode is then loaded by BaseTest.sol
```
QGarchery, line 5:
```suggestion
// Force foundry to compile Morpho Blue even though it's not imported by Metamorpho or by the tests.
// Morpho Blue will be compiled with its own solidity version.
// The resulting bytecode is then loaded by BaseTest.sol.
```
[optional]

QGarchery (commented):
Nice !

test/forge/helpers/BaseTest.sol
```diff
@@ -52,8 +52,7 @@ contract BaseTest is Test {
     address internal MORPHO_OWNER = makeAddr("MorphoOwner");
     address internal MORPHO_FEE_RECIPIENT = makeAddr("MorphoFeeRecipient");
 
-    IMorpho internal morpho =
-        IMorpho(deployCode("lib/morpho-blue/out/Morpho.sol/Morpho.json", abi.encode(MORPHO_OWNER)));
+    IMorpho internal morpho = IMorpho(deployCode("out/Morpho.sol/Morpho.json", abi.encode(MORPHO_OWNER)));
```
Rubilmax, line 55:
```suggestion
    IMorpho internal morpho = IMorpho(deployCode("Morpho", abi.encode(MORPHO_OWNER)));
```

should work too

src/mocks/MorphoImport.sol
Rubilmax: I don't get what hardhat does manually and what we should do

src/mocks/MorphoImport.sol
adhusson: Do you see any advantage to this? My reason for putting `>=0.0.0` so that only `Morpho.sol` specifies the version (if it changes, there's no need to modify `MorphoImport.sol`), but maybe there's a tradeoff I missed.

test/forge/helpers/BaseTest.sol
adhusson: 8d2a5b955f4225ff484c6ecd3b0ced562d8ce624

src/mocks/MorphoImport.sol
Rubilmax: Ah right I get it, it's related to: https://github.com/morpho-org/metamorpho/pull/388#discussion_r1445082587

See the hardhat CI error. We need to specify 0.8.19 as a version used to compile in hardhat config

src/mocks/MorphoImport.sol
Rubilmax: Ah right I get it, see the hardhat CI error. We need to specify 0.8.19 as a version used to compile in hardhat config

MerlinEgalite (approved):

adhusson:
Updated hardhat config by adding a compiler. Hardhat will select the highest compiler version compatible with the pragma in each instance.

FYI you could also be more specific and the following:
```
overrides: {
      "src/mocks/MorphoImport.sol": {
          version: "0.8.19",
          settings: ...
      }
}
```

hardhat.config.ts
```diff
@@ -45,6 +45,15 @@ BigInt.prototype.toSharesDown = function (totalAssets: BigNumberish, totalShares
   return toSharesDown(this as bigint, totalAssets, totalShares);
 };
 
+const commonCompilerSettings : any = {
+  optimizer: { 
+    enabled: true,
+    runs: 200,
+  },
+  viaIR: true,
+  evmVersion: "berlin",
```
QGarchery, line 54: I think it's "paris" for both

hardhat.config.ts
```diff
@@ -45,6 +45,15 @@ BigInt.prototype.toSharesDown = function (totalAssets: BigNumberish, totalShares
   return toSharesDown(this as bigint, totalAssets, totalShares);
 };
 
+const commonCompilerSettings : any = {
+  optimizer: { 
+    enabled: true,
+    runs: 200,
```
QGarchery, line 51: For Morpho Blue, we use 999999 optimizer runs

QGarchery (changes requested):

hardhat.config.ts
```diff
@@ -45,6 +45,15 @@ BigInt.prototype.toSharesDown = function (totalAssets: BigNumberish, totalShares
   return toSharesDown(this as bigint, totalAssets, totalShares);
 };
 
+const commonCompilerSettings : any = {
```
Rubilmax, line 48: bad formatting, run prettier! :D

hardhat.config.ts
QGarchery: We don't care because we trust the compiler so we can assume it's functionally the same, but hardhat runs gas tests so I think we care here

Rubilmax: Fine!

Rubilmax: @adhusson this discussion got resolved but no changes were pushed. I think @QGarchery suggested to not use the same settings between 0.8.21 and 0.8.19

adhusson: I misunderstood my bad, thanks for spotting this. Pushed 880db7f7b074c5f7a88c6583180c01bcc3d43096.

QGarchery (approved):

Rubilmax (approved):

Jean-Grimal (approved):

MathisGD (commented):
LGTM

Issue #333, opened 2023-11-28
Formal verification wish list


QGarchery:
Here is a list of properties that we can try to verify using Certora:
- [x] there is a tree of roles: owner > curator > allocator, and owner > guardian (started in #334, finished in #389)
- [x] depositing puts all the tokens into Morpho Blue (done in #382)
- [x] withdrawing takes all the tokens from Morpho Blue (done in #382)
- [x] reallocating does not change where the tokens are (done in #382)
- [x] ~~cannot have more assets than the cap right after a transaction, except for donations and interest accrual~~ (tried in #419 but abandoned because of tough timeouts)
- [x] timelock is always between the bounds (done in #385)
- [x] fee is always smaller than max fee (done in #385)
- [x] supply queue and withdraw queue have length <= max length (done in #385)
- [x] no dirty pending values (done in #387)
- [x] pending value is different from current value (done in #387)
- [x] withdraw queue has no duplicates (done in #393)
- [x] the enabled field tracks if a market is in the withdraw queue (done in #393)
- [x] a market can't be marked for removal and (have supply cap > 0 or have a pending cap) (done in #399)
- [x] cannot change timelocked variables for a given amount of time (done in #409, see the corresponding computations)
- [x] withdraw queue contains the markets with supply cap > 0 (done in #390)
- [x] input validation and revert condition (done in #390)
- [x] cannot lose fees to unset fee recipient (done in #390)
- [x] cannot set the same value (done in #390)
- [x] prove immutability (done in #398)
- [x] reentrancy safety (done in #396)
- [x] ~~hard liveness properties can withdraw all under conditions, can reallocate all supply from a market under conditions~~
- [x] can pause supply, can recover from takeovers, can remove market even if it reverts (done in #400)
- [x] can change timelocked values in the specified computation time (done in #409)
- [x] additional reallocate revert conditions have been checked separately: it also reverts if a given market isn't enabled (checked in `MarketInteractions.spec`), and it reverts if supplied!=withdrawn (checked in `Tokens.spec`)
- [x] cannot deposit nor withdraw on a non enabled market (done in #410)
- [x] ~~for any tx: the share price, `totalAssets() / totalSupply()`, does not decrease. This is true only assuming that underlying markets are not updated, otherwise there could be some bad debt realization or forced market removal which would break the property~~
- [x] cannot bypass guardian if the guardian is responsive (tackled in #464)
[...]

PR #390, opened 2024-01-10
[Certora] Reverts

QGarchery:
This PRs goal is to verify the condition in which inputs are not valid. In certain case the rule is called `*RevertCondition` instead of `*InputValidation` because we are able to specify all the revert conditions simply


certora/specs/Reverts.spec
```diff
@@ -0,0 +1,313 @@
...
+    submitCap@withrevert(e, marketParams, newSupplyCap);
+
+    assert e.msg.value != 0 ||
+           !hasCuratorRole ||
+           marketParams.loanToken != asset ||
+           lastUpdate == 0 ||
+           pendingCapValidAt != 0 ||
+           removableAt != 0 ||
+           newSupplyCap == assert_uint256(supplyCap)
+        => lastReverted;
```
MathisGD, line 145: same remark

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,313 @@
...
+    uint184 anyCap;
+    anyCap, _, _ = config(newSupplyQueue[i]);
+
+    setSupplyQueue@withrevert(e, newSupplyQueue);
+
+    assert e.msg.value != 0 ||
+           !hasAllocatorRole ||
+           newSupplyQueue.length > maxQueueLength ||
+           anyCap == 0
+        => lastReverted;
```
MathisGD, line 189: why could it fail otherwise?

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,313 @@
...
+    uint256 oldLength = withdrawQueueLength();
+    uint256 anyOtherIndex = indexes[j];
+
+    updateWithdrawQueue@withrevert(e, indexes);
+
+    assert e.msg.value != 0 ||
+           !hasAllocatorRole ||
+           anyIndex > oldLength ||
+           (i != j && anyOtherIndex == anyIndex)
+        => lastReverted;
```
MathisGD, line 208: why could it fail otherwise?

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,313 @@
...
+}
+
+rule reallocateInputValidation(env e, MetaMorphoHarness.MarketAllocation[] allocations) {
+    bool hasAllocatorRole = hasAllocatorRole(e.msg.sender);
+
+    reallocate@withrevert(e, allocations);
+
+    assert e.msg.value != 0 ||
+           !hasAllocatorRole
+        => lastReverted;
```
MathisGD, line 218: - also reverts if a given market isn't enabled
- could we check that it reverts if supplied!=withdrawn

certora/specs/Reverts.spec
QGarchery: I added a comment: it cannot revert for another reason, but we cannot show that because we verify `forall i. cap[i] = 0 => revert` and not `(forall i. cap[i] = 0) => revert`. The formula `forall i. revert => cap[i] = 0` is false, or, said otherwise: it can revert because of any index, but we cannot know the index in advance.

certora/specs/Reverts.spec
QGarchery: I added the other ways it can revert in the comment

MathisGD (approved):
LGTM, let's keep my remark in an issue

certora/specs/Reverts.spec
QGarchery: Added in #333

Rubilmax (approved):

PR #392, opened 2024-01-11
Add multicall encoder

### LINKED ISSUE #321: create encoder for MetaMorpho multicall

Rubilmax:
Fixes #321 

@julien-devatom i need you to create the `npm` environment on this repository pls!


julien-devatom (approved):

julien-devatom (approved):
Environment set up 🔥

pkg/MetaMorphoAction.ts
QGarchery, line 1: It wouldn't cost much to add the main functions to deposit and withdraw, even though I agree that the use case is unclear

QGarchery (approved):
LGTM

Maybe some tests could be added. If I understand correctly the risk of badly encoding the calls is limited because of the static typing, but it's still possible (switch 2 arguments of the same type for example) so it would be nice to have

Rubilmax:
> LGTM
> 
> Maybe some tests could be added. If I understand correctly the risk of badly encoding the calls is limited because of the static typing, but it's still possible (switch 2 arguments of the same type for example) so it would be nice to have

The main benefit is being able to easily batch calls, like it is done in the setup of hardhat tests
Static typing is something, but as you mentioned you can still switch up addresses as input

MerlinEgalite (approved):

PR #393, opened 2024-01-12
[Certora] Enabled flag

QGarchery:
This PR goals are to: 
1. check that there are no duplicate in the withdraw queue
2. check that the `enabled` flag correctly tracks if a market is in the withdraw queue


certora/applyMunging.patch
```diff
@@ -114,7 +114,20 @@ diff -ruN MetaMorpho.sol MetaMorpho.sol
...
-@@ -126,7 +122,7 @@
+@@ -107,6 +103,12 @@
+     /// @inheritdoc IMetaMorphoBase
+     uint256 public lastTotalAssets;
+ 
++    // HARNESS
++    mapping(Id => uint256) public withdrawRank;
++
++    // HARNESS
++    mapping(Id => uint256) public deletedBy;
```
MathisGD, line 125:
```suggestion
+    mapping(Id => uint256) public deletedAt;
```

certora/specs/ConsistentState.spec
```diff
@@ -155,5 +170,92 @@ function hasSupplyCapIsEnabled(MetaMorphoHarness.Id id) returns bool {
...
+
+rule inWithdrawQueueIsEnabledPreservedUpdateWithdrawQueue(env e, uint256 i, uint256[] indexes) {
+    uint256 j;
+    require isInWithdrawQueueIsEnabled(indexes[i]);
+
+    requireInvariant distinctIdentifiers(indexes[i], j);
+
+    updateWithdrawQueue(e, indexes);
+
+    MetaMorphoHarness.Id id = withdrawQueue(i);
```
MathisGD, line 216: could we take that before (I find it clearer)
```suggestion
    MetaMorphoHarness.Id id = withdrawQueue(indexes[i]);
```

certora/specs/ConsistentState.spec
```diff
@@ -155,5 +170,92 @@ function hasSupplyCapIsEnabled(MetaMorphoHarness.Id id) returns bool {
...
+    // Safe require because j is not otherwise constrained.
+    // The ghost variable deletedBy is useful to make sure that markets are not permuted and deleted at the same time in updateWithdrawQueue.
+    require j == deletedBy(id);
```
MathisGD, line 219: I don't get why we need this and why we even need j

certora/applyMunging.patch
```diff
@@ -123,3 +136,34 @@ diff -ruN MetaMorpho.sol MetaMorpho.sol
          DECIMALS_OFFSET = uint8(uint256(18).zeroFloorSub(IERC20Metadata(_asset).decimals()));
  
          _checkTimelockBounds(initialTimelock);
+@@ -338,6 +340,9 @@
+             seen[prevIndex] = true;
+ 
+             newWithdrawQueue[i] = id;
++
++            // HARNESS
++            withdrawRank[id] = i + 1;
```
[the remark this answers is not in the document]
QGarchery: If we use `withdrawIndex` instead of `withdrawRank`, then we can't state the withdraw index correctness property for the index 0. This is because all values are 0 by default in a mapping, so you can't say if it's 0 because `id` is indeed at index 0, or if it is 0 by default. Notably, in the `enabledIsInWithdrawQueue` rule, your witness would be `withdrawIndex(id)` directly, but the prover would give you a counterexample if it returns 0

certora/specs/ConsistentState.spec
```diff
@@ -155,5 +170,92 @@ function hasSupplyCapIsEnabled(MetaMorphoHarness.Id id) returns bool {
...
+function hasDistinctIdentifiers(uint256 i, uint256 j) returns bool {
+    return i != j => withdrawQueue(i) != withdrawQueue(j);
+}
+
+// Check that there are no duplicate markets in the withdraw queue.
+invariant distinctIdentifiers(uint256 i, uint256 j)
+    hasDistinctIdentifiers(i, j)
+{
+    preserved updateWithdrawQueue(uint256[] indexes) with (env e) {
+        require hasDistinctIdentifiers(indexes[i], indexes[j]);
```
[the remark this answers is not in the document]
QGarchery: This is the induction hypothesis, we don't need to verify it. I made that more clear by changing it into a `requireInvariant`

certora/specs/ConsistentState.spec
```diff
@@ -155,5 +170,92 @@ function hasSupplyCapIsEnabled(MetaMorphoHarness.Id id) returns bool {
...
+    isInWithdrawQueueIsEnabled(i)
+filtered {
+    f -> f.selector != sig:updateWithdrawQueue(uint256[]).selector
+}
+
+rule inWithdrawQueueIsEnabledPreservedUpdateWithdrawQueue(env e, uint256 i, uint256[] indexes) {
+    uint256 j;
+    require isInWithdrawQueueIsEnabled(indexes[i]);
+
+    requireInvariant distinctIdentifiers(indexes[i], j);
```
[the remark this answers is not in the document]
QGarchery: It is, and this is something I did not think about before the prover showed me a counterexample: if you have duplicates in the withdraw queue, then you can remove one of them and it will delete the config for the other in the withdraw queue. This means that the enabled flag can be set to `false` by another iteration in the `updateWithdrawQueue` function, but for the same `id`

certora/specs/ConsistentState.spec
```diff
@@ -155,5 +170,92 @@ function hasSupplyCapIsEnabled(MetaMorphoHarness.Id id) returns bool {
...
+
+    return enabled => rank > 0;
+}
+
+// Checks that enabled markets have a positive withdraw rank, according to the withdrawRank ghost variable.
+invariant enabledHasPositiveRank(MetaMorphoHarness.Id id)
+    isEnabledHasPositiveRank(id);
+
+// Check that enabled markets are in the withdraw queue.
+rule enabledIsInWithdrawQueue(MetaMorphoHarness.Id id) {
```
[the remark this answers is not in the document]
QGarchery: This PR's initial goal was to show that the `enabled` flag correctly tracks identifiers that are in the withdraw queue. This rule is one part, and the invariant `inWithdrawQueueIsEnabled` is the other

certora/specs/ConsistentState.spec
QGarchery: Yes we could, but the meaning of `id` is the identifier that we check the property on at the end (with the `assert`), so I don't think it will make that clearer

QGarchery: Keeping it like that makes it clearer that `id`is the same as the one defined in the function `isInWithdrawQueueIsEnabled`

certora/specs/ConsistentState.spec
QGarchery: `j` is here to make sure that there are no duplicates of `id` in the withdraw queue that are removed (because that would mean setting the `enabled` flag to `false` which breaks the property). We know that there are no duplicates at all in the withdraw queue, thanks to the invariant `distinctIdentifiers`, but we can't use `forall` here so we must specify which indexes lead to different identifiers.

The trick is the following:
- we want to specify that there are no identifier that is a duplicate of `id` and that is removed
- if there is one duplicate (or more) that is removed, then then `j` would be an index that correspond to such an identifier in the initial withdraw queue
- we specify that `j` and `indexes[i]` correspond to different identifiers in the initial withdraw queue, which is what we want to specify because `indexes[i]` corresponds to `id` in the initial withdraw queue

certora/applyMunging.patch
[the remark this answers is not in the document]
MathisGD: get it

certora/specs/ConsistentState.spec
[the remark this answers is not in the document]
MathisGD: indeed, nice one

certora/specs/ConsistentState.spec
MathisGD: get it

certora/specs/ConsistentState.spec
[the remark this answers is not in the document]
MathisGD: which induction sorry? what i understand here is that we want to prove that hasDistinctIdentifier is an invariant, but we assume that updateWithdrawQueue (which is touching the identifiers) doesn't break this invariant.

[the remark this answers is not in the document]
QGarchery: An invariant is a property that is always true. It is proven by induction: prove that it's true at deployment (base case), and prove that each function does not change that invariant (induction step). The `preserved` block is allowing us to evaluate some Certora code **before** the corresponding call (here `updateWithdrawQueue`). Since this is an induction step, it's ok to assume that the property is true before doing the call. Here it's useful because the prover would not have guessed by itself that the arguments it needs to pass are `indexes[i]` and `indexes[j]`, by default it tries with the same parameters, that is `i` and `j` directly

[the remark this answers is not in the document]
MathisGD: ok, I didn't remember exactly how to use "preserved blocks". so it's also proved for all functions at the same time right (but not with this precise assumption, just the general assumption)?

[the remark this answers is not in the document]
QGarchery: Yes it proves for all the other functions without issues with the general assumption `hasDistinctIdentifiers(i, j)`. For `updateWithdrawQueue` the prover needed some extra help, and this preserved block only applies to that case

certora/applyMunging.patch
```diff
@@ -114,7 +114,20 @@ diff -ruN MetaMorpho.sol MetaMorpho.sol
...
-@@ -126,7 +122,7 @@
+@@ -107,6 +103,12 @@
+     /// @inheritdoc IMetaMorphoBase
+     uint256 public lastTotalAssets;
+ 
++    // HARNESS
++    mapping(Id => uint256) public withdrawRank;
++
++    // HARNESS
++    mapping(Id => uint256) public deletedAt;
```
MathisGD, line 125: let's add little "natspec" here, as their exact value isn't so clear from the name

certora/specs/DistinctIdentifiers.spec
MathisGD, line 1: why did you extracted this? It's used only in Enabled.spec right?

certora/specs/Enabled.spec
```diff
@@ -0,0 +1,75 @@
...
+    require isInWithdrawQueueIsEnabled(indexes[i]);
+
+    // Safe require because it is a verified invariant.
+    require hasDistinctIdentifiers(indexes[i], j);
+
+    updateWithdrawQueue(e, indexes);
+
+    MetaMorphoHarness.Id id = withdrawQueue(i);
+    // Safe require because j is not otherwise constrained.
+    // The ghost variable deletedAt is useful to make sure that markets are not permuted and deleted at the same time in updateWithdrawQueue.
```
MathisGD, line 32: it's technically impossible to do this right? So it's just a step to help the verifier realize this property and be able to prove the rule?

certora/specs/Range.spec
MathisGD, line 1: same question. Is it to make the verif faster (because it only parallelizes on different files)?

MathisGD (approved):

certora/specs/DistinctIdentifiers.spec
QGarchery: It's only for performance reasons, because otherwise it sometimes times out unexpectedly

certora/specs/Enabled.spec
QGarchery: Yes it's technically impossible because there are no duplicate Id in the withdraw queue. "Explaining" this to the prover is precisely why we introduce this `deletedAt` ghost variable. So yes it's a step to help the verifier (although my guess is that it would not be able to find this intermediary step by itself in a million years)

MathisGD (approved):

MerlinEgalite (approved):

PR #395, opened 2024-01-18
test(hardhat): use pyramidal deposits/withdraws

Rubilmax:
So the idea is that previously:
- allocation was meant to be all equal on non idle markets, but it wasn't because the idle assets were not reallocated
- withdrawing was rarely hitting more than 1 market
- users were only fresh depositing, not re-depositing

What I did is:
- fix the reallocation bug
- make so that each user interacts twice, first time they deposit a lot and withdraw not much, second time they deposit not much and withdraw a lot (with avg total deposits/withdraw equal for each user). It makes so that when interacting the second time, users are on avg withdrawing from the vault and it at least withdraws from all markets for the last user - so it's closer to prod

### DIFF
--- a/test/hardhat/MetaMorpho.spec.ts
+++ b/test/hardhat/MetaMorpho.spec.ts
@@ -215,7 +215,7 @@ describe("MetaMorpho", () => {
       MetaMorphoAction.setFee(BigInt.WAD / 10n),
     ]);
 
-    supplyCap = (BigInt.WAD * 20n * toBigInt(suppliers.length)) / toBigInt(nbMarkets);
+    supplyCap = (BigInt.WAD * 50n * toBigInt(suppliers.length * 2)) / toBigInt(nbMarkets);
 
     const supplyCapData: string[] = [];
     for (const marketParams of allMarketParams) supplyCapData.push(MetaMorphoAction.submitCap(marketParams, supplyCap));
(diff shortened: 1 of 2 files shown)


MerlinEgalite:
what's the motivation behind it?

Rubilmax:
> what's the motivation behind it?

I updated the PR description!

MerlinEgalite (approved):

QGarchery (approved):

test/hardhat/MetaMorpho.spec.ts
```diff
@@ -250,19 +250,35 @@ describe("MetaMorpho", () => {
...
-      logProgress("main", i, suppliers.length);
+    const nbSuppliers = suppliers.length;
+    const nbDeposits = nbSuppliers * 2;
 
-      const supplier = suppliers[i];
-      const assets = BigInt.WAD * toBigInt(1 + Math.floor(random() * 100));
+    for (let i = 0; i < nbDeposits; ++i) {
+      logProgress("main", i, nbDeposits);
+
+      const j = i >= nbSuppliers ? nbDeposits - i - 1 : i;
```
MathisGD, line 259: Not exactly the same order, but more sexy 
```suggestion
      const j = i % nbSuppliers;
```

MathisGD (approved):

PR #396, opened 2024-01-19
[Certora] Reentrancy

QGarchery:
This PR's goals is to prove that there is no possibility of a reentrancy attacks, assuming that:
- Tokens are trusted. This assumption could potentially be relaxed, based on the fact that the accounting does not depend on reentrant calls. In fact, MetaMorpho has been built with this concern in mind, notably to take into account ERC777 tokens. To simplify the verification though, tokens are assumed to not reenter the contract.
- Morpho Blue and Morpho Blue markets are trusted. Morpho Blue is an immutable contract, and thoroughly verified.

Note also that both assumption make sense in the context of timelocked markets: a market (and its corresponding collateral and loan token) can only be added after going through a timelock phase. After a proper analysis during this period, a market could be considered safe

### DIFF
--- a/certora/specs/Reentrancy.spec
+++ b/certora/specs/Reentrancy.spec
@@ -33,7 +33,6 @@ function boolSummary() returns bool {
     return value;
 }
 
-persistent ghost bool delegateCall;
 persistent ghost bool ignoredCall;
 // True when storage has been accessed with either a SSTORE or a SLOAD.
 persistent ghost bool hasAccessedStorage;
@@ -61,10 +60,6 @@ hook CALL(uint g, address addr, uint value, uint argsOffset, uint argsLength, ui
     }
 }
 
-hook DELEGATECALL(uint g, address addr, uint argsOffset, uint argsLength, uint retOffset, uint retLength) uint rc {
-    delegateCall = true;
-}
-
 // Check that no function is accessing storage, then making an external CALL, and accessing storage again.
 rule reentrancySafe(method f, env e, calldataarg data) {
     // Set up the initial state.
@@ -72,12 +67,4 @@ rule reentrancySafe(method f, env e, calldataarg data) {
     require !hasAccessedStorage && !hasCallAfterAccessingStorage && !hasReentrancyUnsafeCall;
     f(e,data);
     assert !hasReentrancyUnsafeCall;
-}
-
-// Check that the contract is truly immutable.
-rule noDelegateCalls(method f, env e, calldataarg data) {
-    // Set up the initial state.
-    require !delegateCall;
-    f(e,data);
-    assert !delegateCall;
 }
--- a/certora/specs/Reentrancy.spec
+++ b/certora/specs/Reentrancy.spec
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+methods {
+    function multicall(bytes[]) external returns(bytes[]) => NONDET DELETE;
+}
+
+ghost bool delegateCall;
+ghost bool callToMorpho;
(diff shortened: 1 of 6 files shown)


certora/specs/Reentrancy.spec
```diff
@@ -0,0 +1,55 @@
...
+methods {
+    function multicall(bytes[]) external returns(bytes[]) => NONDET DELETE;
+
+    function _.accrueInterest(MetaMorphoHarness.MarketParams) external => voidSummary() expect void;
+    function _.supply(MetaMorphoHarness.MarketParams, uint256, uint256, address, bytes) external => uintPairSummary() expect (uint256, uint256);
+    function _.withdraw(MetaMorphoHarness.MarketParams, uint256, uint256, address, address) external => uintPairSummary() expect (uint256, uint256);
+
+    function _.transfer(address, uint256) external => boolSummary() expect bool;
+    function _.transferFrom(address, address, uint256) external => boolSummary() expect bool;
+    function _.balanceOf(address) external => uintSummary() expect uint256;
```
MathisGD, line 11: would it be worth having comments here saying what we expect from external contracts (erc20 isn't reentrant, morpho isn't reentrant)?

MathisGD: actually it's written later
```
// Ignore calls to tokens and Morpho markets as they are trusted (they have gone through a timelock).
```

certora/specs/Reentrancy.spec
```diff
@@ -0,0 +1,55 @@
...
+}
+
+function boolSummary() returns bool {
+    ignoredCall = true;
+    bool value;
+    return value;
+}
+
+persistent ghost bool ignoredCall;
+persistent ghost bool hasCall;
```
MathisGD, line 38: I think that we could have only one "unsafeCall", which is turned true on all calls and turned back to false by the summaries of trusted contracts, but I don't know if it's really better

MathisGD: ah no actually the CALL hook is "executed" after the summary?

MerlinEgalite (approved):

certora/specs/Reentrancy.spec
QGarchery: Yes it is executed after the summary
Screenshot_2024-02-02_11-03-12

certora/specs/Reentrancy.spec
```diff
@@ -0,0 +1,55 @@
...
+
+hook CALL(uint g, address addr, uint value, uint argsOffset, uint argsLength, uint retOffset, uint retLength) uint rc {
+    if (ignoredCall) {
+        // Ignore calls to tokens and Morpho markets as they are trusted (they have gone through a timelock).
+        ignoredCall = false;
+    } else {
+        hasCall = true;
+    }
```
MathisGD, line 46: useless suggestion ^^
```suggestion
    // Ignore calls to tokens and Morpho markets as they are trusted (they have gone through a timelock).
    hasCall = !ignoredCall;
    ignoredCall = false;
```

MathisGD (approved):

certora/specs/Reentrancy.spec
QGarchery: I think you can't do that because otherwise having an ignored call after a non-ignored call will not be caught

MathisGD: nice catch
```suggestion
    // Ignore calls to tokens and Morpho markets as they are trusted (they have gone through a timelock).
    if (!ignoredCall) {
        hasCall = true;  
    }
    ignoredCall = false;
```

QGarchery: It would work too, merging as is for now

PR #401, opened 2024-02-12
use deployCode in IntegrationTest so solidity version is not fixed

adhusson:
MetaMorpho uses Solidity 0.8.21 (strictly). The goal is to have 0.8.24 in the [public allocator](https://github.com/morpho-org/public-allocator) but still be able to reuse Metamorpho's useful test contracts. That means not importing `MetaMorpho` in `IntegrationTest`.

This PR adds a MM-deployer function in `IntegrationTest`. It creates Metamorpho instances using `deployCode`. As with Blue, importers of MM must now have a `MetaMorphoImport.sol` file which adds `MetaMorpho` to the compilation plan.

### DIFF
--- a/test/forge/DeploymentTest.sol
+++ b/test/forge/DeploymentTest.sol
@@ -15,9 +15,7 @@ contract DeploymentTest is IntegrationTest {
         vm.assume(address(notToken) != address(vault));
 
         vm.expectRevert();
-        createMetaMorpho(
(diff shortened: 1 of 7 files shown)


MathisGD (approved):

MerlinEgalite (approved):

QGarchery (approved):
Optimistically approving (without checking that this solution actually removes the dependency on 0.8.21)

PR #410, opened 2024-03-21
[Certora] Market interactions

QGarchery:
Checks that every supply and withdraw on Morpho Blue targets an enabled market


certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+    require isWithdrawRankCorrect(id);
+
+    // Check that all markets from which MetaMorpho withdraws are enabled markets.
+    assert config_(id).enabled;
+
+    // NONDET summary, which is sound because all non view functions in Morpho Blue are abstracted away.
+    return (_, _);
+}
+
+// Check assertions in the summaries.
```
MathisGD, line 57: smart

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+// This requires to turn off sanity checks for this invariant that appears vacuous.
```
MathisGD, line 58: I don't understand this, which require?

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+    require Util.libId(marketParams) == id;
+
+    return marketParams;
+}
+
+function summarySupply(MetaMorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns(uint256, uint256) {
+    assert shares == 0;
+    assert onBehalf == currentContract;
+    assert data.length == 0;
+
```
MathisGD, line 27:
```suggestion
    assert assets > 0;
    
```
can also be verified, although that's not what we are trying to do

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+    assert config_(id).enabled;
+
+    // NONDET summary, which is sound because all non view functions in Morpho Blue are abstracted away.
+    return (_, _);
+}
+
+function summaryWithdraw(MetaMorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, address receiver) returns (uint256, uint256) {
+    assert onBehalf == currentContract;
+    assert receiver == currentContract;
+
```
MathisGD, line 42:
```suggestion
    assert assets > 0;
    
```
same

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function summarySupply(MetaMorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns(uint256, uint256) {
+    assert shares == 0;
+    assert onBehalf == currentContract;
+    assert data.length == 0;
+
+    MetaMorphoHarness.Id id = Util.libId(marketParams);
+    // Safe require because it is a verified invariant
+    require hasSupplyCapIsEnabled(id);
```
MathisGD, line 30: it doesn't work without this require?

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+
+    // Check that all markets on which MetaMorpho supplies are enabled markets.
+    assert config_(id).enabled;
+
+    // NONDET summary, which is sound because all non view functions in Morpho Blue are abstracted away.
```
MathisGD, line 35: what do you mean?

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+    return (_, _);
```
MathisGD, line 36: what does this do?

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+function summaryWithdraw(MetaMorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, address receiver) returns (uint256, uint256) {
+    assert onBehalf == currentContract;
+    assert receiver == currentContract;
+
+    MetaMorphoHarness.Id id = Util.libId(marketParams);
+    uint256 index = lastIndexWithdraw();
+    // Safe require because it is a verified invariant.
+    require isInWithdrawQueueIsEnabled(index);
+    // Safe require because it is a verified invariant
+    require isWithdrawRankCorrect(id);
```
MathisGD, line 48: not intuitive why do we need this

certora/specs/MarketInteractions.spec
```diff
@@ -0,0 +1,60 @@
...
+
+    // Check that all markets from which MetaMorpho withdraws are enabled markets.
+    assert config_(id).enabled;
+
+    // NONDET summary, which is sound because all non view functions in Morpho Blue are abstracted away.
+    return (_, _);
```
MathisGD, line 54: same questions

certora/specs/MarketInteractions.spec
QGarchery: I should reword because it's not a `require` but I was saying that we needed to turn off sanity checks because otherwise the verification will fail. So that's why the line `"rule_sanity": "basic"` does not appear in the configuration file

MathisGD: Why does the sanity check fail

certora/specs/MarketInteractions.spec
QGarchery: No, because we don't check that it is enabled in the code, we check that the supply cap is > 0

certora/specs/MarketInteractions.spec
QGarchery: It should not be needed, I forgot to remove it. The idea before was to use `withdrawRank` and subtract 1 to have the `index`, so that's why this invariant would have been useful. In the end I couldn't prove that `withdrawRank` was greater than 1 so I used a ghost variable instead

certora/specs/MarketInteractions.spec
QGarchery: I suspect it could be difficult to prove, since there are case where it withdraws by shares. Let me try to do it

certora/specs/MarketInteractions.spec
QGarchery: It's like if it returned a random value. This is the same as saying that you don't know anything about the return value. This is the safest option for returned values, but it omits storage change. I think it's fine to summarize it this way because all other functions of Morpho Blue are summarized as being able to do anything (notably on storage) on contracts other than MM, so the fact that this does not modify the storage should not impact the soundness of the verification

certora/specs/MarketInteractions.spec
QGarchery: See [this answer](https://github.com/morpho-org/metamorpho/pull/410/files/af889546e995a89c3a1e27bb521cada28f806e38#r1535955139)

certora/specs/MarketInteractions.spec
QGarchery: Because it sees `true` and it says that the invariant should always hold. But we are not interested in the invariant itself, but about running all the functions, and doing the checks in the summaries when they are called

certora/specs/MarketInteractions.spec
MathisGD:
```suggestion
    assert assets == 0 xor shares == 0;
```

certora/specs/MarketInteractions.spec
MathisGD: >It's like if it returned a random value. This is the same as saying that you don't know anything about the return value. This is the safest option for returned values, 

I get this

>but it omits storage change. I think it's fine to summarize it this way because all other functions of Morpho Blue are summarized as being able to do anything (notably on storage) on contracts other than MM, so the fact that this does not modify the storage should not impact the soundness of the verification

but not this

the problem could be that we don't specify how does this function change the storage of Morpho, and of the other contracts (tokens etc...) right?

but you are saying that it's ok, because the prover usually assumes that MM can change the storage of other contract in any way. But I don't get why it's ok to assume no storage change on MM and Blue (imagine a blue function disables a market on MM).

QGarchery: > the problem could be that we don't specify how does this function change the storage of Morpho, and of the other contracts (tokens etc...) right?
but you are saying that it's ok, because the prover usually assumes that MM can change the storage of other contract in any way. 

Yes exactly. The default summarization is called HAVOC_ECF and it allows changes on the storage of other contracts, but leave the current contract  (MM in this case) unchanged, and return any value. The NONDET summary is to leave every contract unchanged, and return any value. So the difference is in the change of storage of other contracts, but as you said, it should not impact the verification since it already can change arbitrarily for every other function calls.

> But I don't get why it's ok to assume no storage change on MM and Blue (imagine a blue function disables a market on MM).

So this is more about the HAVOC_ECF summarization and why it's sound. One good case is when you only target contracts that were deployed previously and don't allow to call arbitrary other contracts. But this is usually not trivial, and relies on the fact that the contracts you call are trusted. In our case we don't call Blue with callbacks and we verified that there are no untrusted calls, so this summarization should work (but you don't know anything about the details of the functions called)

MathisGD (approved):

MerlinEgalite (approved):

Issue #412, opened 2024-03-28
Proposal for additional verification specs


daejunpark:
In addition to the [existing specs](https://github.com/morpho-org/metamorpho/tree/398c553b06769fc3a84d044c4d1f097de77a8548/certora), the following specs could also be considered. If the prover is incapable of verifying them, fuzzing could be considered as an alternative.

1. **[Share Price Preservation]:** For any tx: the share price, `totalAssets() / totalSupply()`, does _not_ decrease.
   - That is, the price immediately before the tx must be less than or equal to the price immediately after the tx.
   - Note: this spec ensures the correctness of accounting logic, e.g., interest and fees are accurately accounted for, the share price is properly updated prior to converting asset amounts to/from shares, etc.

1. **[ERC4626]:** The ERC4626 properties, especially the [round-trip properties](https://github.com/a16z/erc4626-tests/blob/main/ERC4626.prop.sol#L244-L318), hold.

1. **[Consistency of lastTotalAssets]:** After any tx: `lastTotalAssets == totalAssets()`
   - Small discrepancies may exist due to rounding errors.

1. **[Soft Supply Cap]:** For any tx, for any market: if the supply increases, then it must _not_ exceed the supply cap.
   - Note: the supply cap may be updated to be below the current supply, thus the supply may exceed the cap temporarily. However, it is not allowed to add further supply once it exceeds the cap. Moreover, the excessive supply may be partially withdrawn, where the remaining supply may still exceed the cap.

1. **[Consistency between supplyQueue and withdrawQueue]:** If "`set(supplyQueue) - set(withdrawQueue)`" is non-emtpy, then their cap is zero, with no pending cap.
   - Note: ideally, supplyQueue should be a subset of withdrawQueue. However, a market may be removed from withdrawQueue without updating supplyQueue. That's why certain markets may appear only in supplyQueue. However, the supply cap of such markets must be zero, because of the condition for removal from withdrawQueue.

1. **[Consistency of withdrawQueue]:** A market whose cap > 0 must be included in the withdraw queue.
   - Note: this is the converse of the existing spec [`inWithdrawQueueIsEnabled`](https://github.com/morpho-org/metamorpho/blob/398c553b06769fc3a84d044c4d1f097de77a8548/certora/specs/Enabled.spec#L13).

1. **[Consistency of Market Config]:** For any market config:
   - If it's _not_ enabled, then cap == 0 (although a pending cap may exist). That is, if cap > 0, then it's enabled.
[...]

PR #413, opened 2024-04-02
[Certora] Implement reviews suggestions

QGarchery:
This PR:
- implements last point of #412 
- uses the canonical pattern of checking timestamp properties in `Timelock.spec`

### DIFF
--- a/certora/specs/Timelock.spec
+++ b/certora/specs/Timelock.spec
@@ -13,7 +13,7 @@ function summarySupplyshares(MetaMorphoHarness.Id id, address user) returns uint
     return res;
 }
 
-ghost uint256 lastTimestamp;
+persistent ghost uint256 lastTimestamp;
 
 hook TIMESTAMP uint v {
     // Safe require because timestamps are guaranteed to be increasing.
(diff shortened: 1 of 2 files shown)


certora/specs/Timelock.spec
```diff
@@ -13,13 +13,21 @@ function summarySupplyshares(MetaMorphoHarness.Id id, address user) returns uint
     return res;
 }
 
+persistent ghost uint256 lastTimestamp;
+
+hook TIMESTAMP uint newTimestamp {
```
MathisGD, line 18: I get the concept, but when exactly is that "executed"? Because in the EVM, the timestamp is never changed, it's more an externality and the evm then realises that it changed

QGarchery: Yes it's different from the EVM in the sense that this `lastTimestamp` variable tracks all interactions, including view function calls that could be seen as just looking at the chain. Essentially, it allows to have a notion of the "current time", which you can compare to previous times.
For example, if a function call stores the `block.timestamp`, then you know that the stored value is smaller than the current time. This means that for every interactions that happen after (in the CVL spec, modeling interactions that happen after in the real life) will have a `block.timestamp` that is greater, including for view functions

MathisGD (approved):

MerlinEgalite (approved):

PR #419, opened 2024-05-17
[Certora] Supply cap

QGarchery:


QGarchery:
Abandoned, as it seems to be too much for the solvers

PR #437, opened 2025-02-14
Update foundry v1.0

QGarchery:
Also update libs:
- forge-std to latest version, to use expectPartial match
- morpho-blue to latest version (avoid license warning, and add latest test and verif changes)
- erc4626-test as they used the testFail pattern before update

Also remove timestamps in munging patch


colin-morpho (approved):

foundry.toml
MathisGD, line 1: let's remove this below

```
[profile.test]
via-ir = false
```

MathisGD (commented):
This change changes the compiled bytecode (because we still have the metadata in this repo). Should we at least remove the metadata?

foundry.toml
QGarchery: Note that:
- it's useful to have this to run tests quickly, this is used by the `yarn test:forge` command
- tests don't run without it, because of [this problem](https://morpholabs.slack.com/archives/C02N7CZ088N/p1739798395868139?thread_ts=1726749618.200599&cid=C02N7CZ088N) with the difference between classic and via-ir semantics

test/forge/MetaMorphoInternalTest.sol
adhusson, line 1: I'm getting this in the `setUp`

```
Encountered 1 failing test in test/forge/MetaMorphoInternalTest.sol:MetaMorphoInternalTest
[FAIL: OwnableInvalidOwner(0x0000000000000000000000000000000000000000)] setUp() (gas: 0)
```

I'm guessing it related to via_ir.

QGarchery: Referencing this in https://github.com/morpho-org/metamorpho/issues/438. For now I think we can merge this PR to recover the CI as it's more urgent

foundry.toml
```diff
@@ -4,9 +4,12 @@ test = "test/forge"
 libs = ["lib"]
 evm_version = "paris"
 via-ir = true
+optimizer = true
+optimizer_runs = 200
+bytecode_hash = "none"
 fs_permissions = [
-    { access = "read", path = "./config/"},
-    { access = "read", path = "./out/"}
+    { access = "read", path = "./config/" },
```
MathisGD, line 11:
```suggestion
```

MathisGD (approved):
verified that it doesn't modify the bytecode (if we remove the metadata on main as well)

foundry.toml
QGarchery: Do you remember why either read access was granted ? I can't seem to find what breaks if we remove `fs_permission` altogether

QGarchery: Removing optimistically, as it's not hard to reintroduce it

PR #440, opened 2025-03-31
Remove npm package

QGarchery:


package.json
MathisGD, line 1: we might be able to remove some dependancies as well (not sure how to find the useless ones though)

QGarchery: I don't think we can remove dependencies, because the imports (`ethers` and `types`) are already used by the hardhat tests

MathisGD (approved):

colin-morpho (approved):

bmarcot:
The command `yarn add @morpho-org/metamorpho` results in an empty dependency when added to an existing project. Could this be related to the issue? What is the recommended method for adding `metamorpho` to an existing project? I need access to the interface.

MathisGD:
we recommend using git submodules for metamorpho contracts. you might as well be interested in https://github.com/morpho-org/sdks for typescript sdks

PR #445, opened 2025-07-21
Compatibility via-ir

### LINKED ISSUE #438: Make tests compatible with via-ir pipeline
_Originally posted by @adhusson in https://github.com/morpho-org/metamorpho/pull/437#discussion_r1967394900_
      
Ideally we would only have to do `forge t`

QGarchery:
This PR:
- fixes #438 
- does so by aligning the testing to how it's done in MMv1.1 (use a mock) so the difference with MMv1.1 is reduced following this PR
- also renames `test` profile to `no_via_ir` to avoid the confusion that `forge test` does not necessarily use the `test` profile

See https://morpholabs.slack.com/archives/C02N7CZ088N/p1752855061376499?thread_ts=1726749618.200599&cid=C02N7CZ088N. There seems to also be a semantic change where constructor arguments are evaluated first in the via-ir pipeline


test/forge/helpers/InternalTest.sol
```diff
@@ -2,17 +2,41 @@
...
 
 import "./BaseTest.sol";
-import {MetaMorpho} from "../../../src/MetaMorpho.sol";
+import {MetaMorpho, SafeERC20} from "../../../src/MetaMorpho.sol";
 
 contract InternalTest is BaseTest, MetaMorpho {
+    using SafeERC20 for IERC20;
+
+    address constant tempOwner = address(100);
+    address constant tempMorpho = address(101);
```
QGarchery, line 11: Needed because you can't use `address(0)`: it would revert.

lib/openzeppelin-contracts
QGarchery, line 1: upgrade needed to be able to override the `asset` function effectively. The precise change needed was the replacement of `_asset` into `asset()` in v5.0.3, see the `ERC4626.sol` file in [this diff](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/v5.2.0...v5.3.0#diff-0a51bef77bfde9f5057fe5a66fbaf19b3e2982059363e752d26a25073ad51bfeR246-R247)

test/forge/helpers/InternalTest.sol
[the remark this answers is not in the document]
QGarchery: Do we want to also test it on the CI ?

adhusson (approved):

src/mocks/MetaMorphoMock.sol
```diff
@@ -0,0 +1,27 @@
...
+contract MetaMorphoMock is MetaMorpho {
+    constructor(
+        address owner,
+        address morpho,
+        uint256 initialTimelock,
+        address _asset,
+        string memory __name,
+        string memory __symbol
+    ) MetaMorpho(owner, morpho, initialTimelock, _asset, __name, __symbol) {}
+
```
MathisGD, line 15:
```suggestion
```

PR #451, opened 2026-04-14
remove yarn

MathisGD:


peyha (approved):
The yarn tests are essentially gas tests, not really worth keeping

.github/workflows/foundry.yml
```diff
@@ -81,8 +81,11 @@ jobs:
 
       - uses: ./.github/actions/install-cache
 
+      - name: Build morpho-blue
+        run: cd lib/morpho-blue && forge build
+
       - name: Run tests in ${{ matrix.type }} mode
-        run: yarn test:forge -vvv
+        run: FOUNDRY_PROFILE=no_via_ir forge test -vvv
```
QGarchery: Either both via ir or both without (previously it was both without)

.github/actions/install/action.yml
```diff
@@ -4,15 +4,5 @@ runs:
...
 
   steps:
-    - name: Install node
-      uses: actions/setup-node@v4
-      with:
-        node-version: 18
-        cache: yarn
-
     - name: Install Foundry
       uses: foundry-rs/foundry-toolchain@v1
```
QGarchery:
```suggestion
      uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1
```
and similar change in `.github/install-cache/action.yml`

.github/workflows/foundry.yml
```diff
@@ -79,10 +55,11 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: ./.github/actions/install-cache
+      - name: Build morpho-blue
+        run: cd lib/morpho-blue && forge build
 
       - name: Run tests in ${{ matrix.type }} mode
-        run: yarn test:forge -vvv
+        run: forge test -vvv
```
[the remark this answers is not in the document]
MathisGD: hmm weird that tests were passing then

QGarchery (approved):

adhusson (commented):
LGTM

PR #463, opened 2026-07-15
Add audit findings comments

QGarchery:
Those were mostly covered by audit reports, but better to add it in the code for visibility.

[corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1782902249138749)

### DIFF
--- a/src/interfaces/IMetaMorpho.sol
+++ b/src/interfaces/IMetaMorpho.sol
@@ -98,9 +98,12 @@ interface IMetaMorphoBase {
     /// @dev Warning: Reverts if a cap is already pending. Revoke the pending cap to overwrite it.
     /// @dev Warning: Reverts if a market removal is pending.
     /// @dev In case the new cap is lower than the current one, the cap is set immediately.
+    /// @dev Market's loan token and IRM should not reenter.
     function submitCap(MarketParams memory marketParams, uint256 newSupplyCap) external;
 
     /// @notice Accepts the pending cap of the market defined by `marketParams`.
+    /// @dev Accepting a market for which the vault has supply works the same way as a donation to the vault.
+    /// In particular, it can be extracted by front-running the transaction.
     function acceptCap(MarketParams memory marketParams) external;
 
     /// @notice Revokes the pending cap of the market defined by `id`.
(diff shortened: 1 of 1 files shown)


src/interfaces/IMetaMorpho.sol
```diff
@@ -98,16 +98,20 @@ interface IMetaMorphoBase {
...
+    /// @dev Accepting a market for which the vault has supply works the same way as a donation to the vault.
+    /// In particular, it can be extracted by front-running the transaction.
     function acceptCap(MarketParams memory marketParams) external;
 
     /// @notice Revokes the pending cap of the market defined by `id`.
     /// @dev Does not revert if there is no pending cap.
     function revokePendingCap(Id id) external;
 
     /// @notice Submits a forced market removal from the vault, eventually losing all funds supplied to the market.
+    /// @notice Notably, adding back a removed market should be considered the same as adding a new market.
```
MathisGD:
```suggestion
    /// @dev Notably, adding back a removed market should be considered the same as adding a new market.
```

src/interfaces/IMetaMorpho.sol
```diff
@@ -98,16 +98,20 @@ interface IMetaMorphoBase {
...
+    /// @dev Accepting a market for which the vault has supply works the same way as a donation to the vault.
+    /// In particular, it can be extracted by front-running the transaction.
     function acceptCap(MarketParams memory marketParams) external;
 
     /// @notice Revokes the pending cap of the market defined by `id`.
     /// @dev Does not revert if there is no pending cap.
     function revokePendingCap(Id id) external;
 
     /// @notice Submits a forced market removal from the vault, eventually losing all funds supplied to the market.
+    /// @dev Notably, adding back a removed market should be considered the same as adding a new market.
```
MathisGD: what do you mean exactly here?

QGarchery: Meant to highlight that adding a market that was force removed doesn't mean that the funds are being recovered. It should be treated the same way as a new market, with no expectation about the vault supply that already exists. Because if you see it like funds being recovered (which doesn't make sense, because the comment above states that the funds are lost), then you could think that there is an issue because those funds could be extracted with a sandwich attack (also they are not recovered to the right set of vault depositors)

MathisGD: ok, I would be slightly more explicit then

QGarchery: added a more explicit sentence about the "recovery"

MathisGD (approved):

peyha (approved):

PR #466, opened 2026-07-30
fix: document maxWithdraw edge-case

MathisGD:


src/MetaMorpho.sol
```diff
@@ -507,13 +507,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.
```
MathisGD:
```suggestion
    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that
    /// _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a
    /// Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.
```

src/MetaMorpho.sol
```diff
@@ -507,13 +507,15 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
     /// @dev Warning: May be lower than the actual amount of assets that can be withdrawn by `owner` due to conversion
     /// roundings between shares and assets.
     function maxWithdraw(address owner) public view override returns (uint256 assets) {
         (assets,,) = _maxWithdraw(owner);
     }
 
     /// @inheritdoc IERC4626
+    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.
```
MathisGD:
```suggestion
    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that
    /// _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a
    /// Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.

```

QGarchery (approved):

src/MetaMorpho.sol
```diff
@@ -507,13 +507,20 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
     }
 
     /// @inheritdoc IERC4626
+    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that
+    /// _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a
+    /// Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.
```
peyha, line 512:
```suggestion
    /// Blue callback, the Blue balance check is guaranteed because it is larger than the liquidity of the withdrawable markets
```
new attempt

peyha (approved):

src/MetaMorpho.sol
```diff
@@ -507,13 +507,20 @@ contract MetaMorpho is ERC4626, ERC20Permit, Ownable2Step, Multicall, IMetaMorph
...
     /// roundings between shares and assets.
     function maxWithdraw(address owner) public view override returns (uint256 assets) {
         (assets,,) = _maxWithdraw(owner);
     }
 
     /// @inheritdoc IERC4626
+    /// @dev May overestimate the real result when called in the middle of a Blue callback execution. The reason is that
+    /// _simulateWithdrawMorpho counts the Blue loan token balance once per market instead of once. But when not in a
+    /// Blue callback, the Blue balance cannot be the limiting factor so it does not overestimate then.
+
```
MathisGD:
```suggestion
```

